PR #2, opened 2019-04-06
Added method to get_pyarrow_filesystem and deleted .gitignore folder

hayesgb:
Added pyarrow filesystem support for the Gen1 Azure Datalake FS Client.  Will be creating a matching PR in the pyarrow project to support this addition.  Matching PR includes implementation of an ADLFSWrapper in arrow/python/pyarrow/filesystem.py


martindurant:
This is fine, but I'd also like to point you to [fsspec](https://github.com/martindurant/filesystem_spec), where any class deriving from `AbstractFileSystem` is also a subclass of the pyarrow `FileSystem` (if pyarrow is installed). adlfs has not been ported to fsspec, but I would argue it should be - or perhaps, given the arrival of the gen2 API, now if the time to develop a separate package.

martindurant:
Please tag back here when the pyarrow PR is complete.

hayesgb:
Attempting to bring both the Azure Gen1 and Gen2 datalake filesystems into a single package.  There are currently no tests associated with the code; I've just started a branch to start working on tests for the Gen2 class.  

The Gen1 filesystem is implemented using multiple inheritance, while the Gen2 filesystem is implemented through the ADL Gen2 API.  Currently, there is no session caching.  Intent would be to get tests in place via mocking the API response, then implement session caching, then begin parsing the url to simplify authentication, and reduce the complexity of authenticating via the storage_options parameter in dask.

Last thing.  As mentioned in the fsspec discussion, here I've added the AllBytes class as a workaround to the apparent bug in BytesCache (with small files).  I can take a look at this as well.

PR #9, opened 2019-11-15
Wip clean

TomAugspurger:


adlfs/core.py
```diff
@@ -170,12 +171,14 @@ def __setstate__(self, state):
...
+    # TODO: refoctor this. I suspect we actually want to compose an
+    # AbstractBufferedFile with an AzureDLFile.
+    def __init__(self, fs, path, mode="rb", block_size=2 ** 25, delimiter=None):
         super().__init__(
             azure=fs.azure_fs,
             path=AzureDLPath(path),
             mode=mode,
-            blocksize=blocksize,
+            blocksize=block_size,
             delimiter=delimiter,
```
martindurant: was cache_type leaking to here in latest fsspec?

TomAugspurger: Fixed, maybe. Need to actually test it though.

PR #14, opened 2019-11-18
pass through cache_options

TomAugspurger:
Closes https://github.com/intake/filesystem_spec/issues/205

@hayesgb can you test this out with fsspec>=0.6? We don't have unit tests for gen1 AFAICT.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 azure-datalake-store>=0.0.46,<0.1
-azure-storage-blob>=2.1.0
+azure-storage-blob>=2.1.0,<3.0.0
 fsspec>=0.6.0,<1.0
 pytest>=5.0,<6.0
 requests>=2.22.0,<3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 azure-datalake-store>=0.0.46,<0.1
 azure-storage-blob>=2.1.0
-fsspec>0.4.0,<1.0
+fsspec>=0.6.0,<1.0
 pytest>=5.0,<6.0
 requests>=2.22.0,<3.0
--- a/adlfs/core.py
+++ b/adlfs/core.py
@@ -181,6 +181,7 @@ class AzureDatalakeFile(AzureDLFile):
         autocommit=True,
         block_size=2 ** 25,
         cache_type="bytes",
+        cache_options=None,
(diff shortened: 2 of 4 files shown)


TomAugspurger:
cc @JoranDox.

hayesgb:
Currently traveling for work, but I should be able to test it tonight.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 azure-datalake-store>=0.0.46,<0.1
 azure-storage-blob>=2.1.0
```
hayesgb: Need to pin this to azure-storage-blob<3.0  Microsoft is currently doing a large rewrite, and the version on PyPi jumped from 2.1.0 --> 12.0.

martindurant: > the version on PyPi jumped from 2.1.0 --> 12.0.

! 
It must be six times better!

TomAugspurger:
In general, do we want to set max pins?

hayesgb:
That’s hilarious @martindurant.  

The only reason I would suggest it is to minimize the risk of breaking changes with major version changes in dependencies.

hayesgb:
This seems to work just fine with fsspec=0.6.0 in Azure Datalake Gen1.

martindurant:
Time to add to fsspec.registry.known_implementations?

TomAugspurger:
Probably. Is that done on the fsspec or adlfs side?

Have you considered using entrypoints for this so that it can be done
without needing to import the package?

martindurant:
in fsspec: https://github.com/intake/filesystem_spec/blob/master/fsspec/registry.py#L12 . Each package is only imported when required.

Yes, entry points would be a reasonable way to do this for the future.

hayesgb:
What was the final decision on abfs as a protocol?  Everything is written for abfs, but an alternative was suggested.

martindurant:
I vote "abs"

hayesgb:
I’m good with either — my only hesitation is whether we should maintain consistency with the Azure HDFS specification, which is abfs[s]://

Thoughts?

martindurant:
Is that format published somewhere? If yes, then I suppose it makes our decision for us.

hayesgb:
It is.  I can send a link to the docs later, but it’s on the Apache site.

PR #17, opened 2019-11-23
Docker tests and packaging

AlbertDeFusco:
@martindurant , @TomAugspurger 

- explicitly support all BlockBlobService init args in AzureBlobFileSystem
- implement automatic testing with `pytest-docker-compose`. The conda package is available [on my channel](https://anaconda.org/defusco/pytest-docker-compose)
- fix a few minor issues with the tests. Each test will initiate the docker container and prepare the Azure container from scratch.
- add versioneer
- add conda recipe
- removed unnecessary required packages


hayesgb:
@AlbertDeFusco -- I was attempting to run pytest locally, and I'm getting an error when I try to install pytest-docker-compose with pip.  It centers on PyYAML, as follows:
`  Found existing installation: PyYAML 5.1.2
ERROR: Cannot uninstall 'PyYAML'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.`

Also - looking at the github repo for this project, it looks like it isn't supported on Python3.7.  I have some reservations about including a dependency in the build pipeline that isn't supported in Python3.7.  Any chance you can take a look at the PyYAML bug and extending support for Python3.7?

adlfs/core.py
```diff
@@ -230,9 +231,46 @@ class AzureBlobFileSystem(AbstractFileSystem):
 
     Parameters
     ----------
-    storage_account:  Name of the Azure Storage Account
-    account_key:  Access key for the Azure Storage account
-    container_name:  Name of the container or filesystem to be accessed (optional)
+    account_name:
+            The storage account name. This is used to authenticate requests
```
TomAugspurger: Typically just indented 4 additional spaces.

adlfs/core.py
```diff
@@ -252,18 +290,32 @@ class AzureBlobFileSystem(AbstractFileSystem):
...
+        is_emulated=False,
+        protocol=DEFAULT_PROTOCOL,
+        endpoint_suffix=SERVICE_HOST_BASE,
+        custom_domain=None,
+        request_session=None,
+        connection_string=None,
+        socket_timeout=None,
+        token_credential=None,
     ):
         AbstractFileSystem.__init__(self)
```
TomAugspurger: It's potentially important to pass through all arguments to `super().__init__()`. The instance caching is based on these arguments.

azure-pipelines.yml
```diff
@@ -4,26 +4,26 @@ trigger:
...
+# services:
+#   azurite: azurite
 
 steps:
 - bash: echo "##vso[task.prependpath]$CONDA/bin"
   displayName: Add conda to PATH
 
 - script: |
-    conda install -y -c conda-forge python=3 pre-commit azure-datalake-store azure-storage-blob fsspec pytest requests
+    conda install -y -c conda-forge -c defusco python=3 pre-commit azure-datalake-store azure-storage-blob fsspec pytest requests pytest-docker-compose
```
TomAugspurger: Any chance you get get this added to conda-forge?

adlfs/core.py
martindurant: Doesn't that happen before init is called at all? I would say, that the signature should contain **kwargs, and we pass those on.

TomAugspurger: Ah, yes you're right. `AbstractFileSystem.__init__` doesn't actually use args or kwargs anymore.

PR #19, opened 2019-11-26
Update README.md

Timost:
Hi !
First of all thank you for working on this much needed azure storage layer.

I think the readme is not correct for the Gen 2 details. I think it should pass the `AzureBlobFileSystem` class instead of the `AzureDatalakeFileSystem`.

If that is correct. Here is the PR fixing the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ To use the Gen2 filesystem:
 ```
 import dask.dataframe as dd
 from fsspec.registry import known_implementations
-known_implementations['abfs'] = {'class': 'adlfs.AzureDatalakeFileSystem'}
+known_implementations['abfs'] = {'class': 'adlfs.AzureBlobFileSystem'}
 STORAGE_OPTIONS={'account_name': ACCOUNT_NAME, 'account_key': ACCOUNT_KEY}
 
 ddf = dd.read_csv('abfs://{CONTAINER}/{FOLDER}/*.csv', storage_options=STORAGE_OPTIONS}
--- a/setup.py
+++ b/setup.py
@@ -21,8 +21,6 @@ setup(
         "azure-storage-blob>=2.1.0,<3.0.0",
         "fsspec>=0.6.0,<1.0",
     ],
-    tests_require=[
-        "pytest>5.0,<6.0",
-    ],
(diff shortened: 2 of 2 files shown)


martindurant:
Thank you for picking up on this.
After the coming change in fsspec, this step should not be necessary at all, so I expect this PR to be superseded ( cc @TomAugspurger )

(note that the failure is due to the implementation of linting)

Timost:
Thanks for the quick reply ! Regarding the upcoming change in fsspec. Do you mean this PR: intake/filesystem_spec#206 ?

martindurant:
That is an issue yet to be implemented, and that would work too, but I actually mean directly adding the implementation classes here to fsspec.registry.known_implementations

Timost:
Oh ok, I couldn't find the related PR.! Are you interested in this PR then ? If yes, do you want me to edit the setup.py file for the linting to pass ?

martindurant:
Yes, that PR to fsspec should be made.
Following that, the lines here concerning the fsspec registry should simply be removed, they will not be necessary. Regarding linting, yes, we should just accept what the linter says is the right format (I don't know why it changed).

hayesgb:
Cross-referencing with this [PR.](https://github.com/intake/filesystem_spec/pull/209)

README.md
```diff
@@ -23,7 +23,7 @@ To use the Gen2 filesystem:
 ```
 import dask.dataframe as dd
 from fsspec.registry import known_implementations
-known_implementations['abfs'] = {'class': 'adlfs.AzureDatalakeFileSystem'}
+known_implementations['abfs'] = {'class': 'adlfs.AzureBlobFileSystem'}
```
martindurant: This shouldn't be needed anymore, right?

hayesgb:
Correct — once the protocols are in the fsspec known_protocols registry, we can remove this from the adlfs Readme.

Timost:
Ok I guess when intake/filesystem_spec#209 is released, this PR can be closed ? Thank you for adding this support !

Timost:
I'm closing this PR. The ffspec `known_protocols` registry no longer needs patching as `adl` and `abfs` are now there in fsspec.

PR #41, opened 2020-02-26
Make DIRECTORY type lower case to be compatible with fsspec walk method

bemgreem:
Modification to the adlfs ls method to fix an inconsistency that breaks fsspec walk method. 

Currently the azure-data-lake-store code uses a file type of 'DIRECTORY' whereas fsspec expects a lower case value in the walk method (specifically  `info["type"] == "directory"` in walk method here https://github.com/intake/filesystem_spec/blob/master/fsspec/spec.py)

This change swaps out the uppercase for lowercase value in adlfs ls method. Tested and walk method works with this change

### DIFF
--- a/adlfs/core.py
+++ b/adlfs/core.py
@@ -99,8 +99,8 @@ class AzureDatalakeFileSystem(AbstractFileSystem):
         )
 
         for file in files:
-            if 'type' in file and file['type'] == 'DIRECTORY':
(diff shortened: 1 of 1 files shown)


hayesgb:
Please lint and re-submit.

martindurant:
Thanks @bemgreem

PR #70, opened 2020-07-06
Fix mkdir when storage account has hierarchical namespace enabled

firecast:
When storage accounts have hierarchical namespace enabled, the paths returned have a trailing `/` in them

Ex. `ls("")` on the storage account will return
```python3
["abc/", "abc2/"]

```

So the condition on the ls fails in that case


hayesgb:
Thanks!

PR #80, opened 2020-08-11
Fix inconsistencies for type and size/length between MS Azure API and fsspec

dfigus:
fsspec is expecting the `type` values `file` and `directory`, however the official MS Azure API is returning upper case values. Hence converting to lower case in `info()` and `ls()` consistently (previously only `directory` was converted).

Furthermore fsspec is expecting `size` instead of `length` used by the MS Azure API. This conversion was already done in `info()` and is now extended to `ls()` with this PR.

### DIFF
--- a/adlfs/core.py
+++ b/adlfs/core.py
@@ -103,8 +103,8 @@ class AzureDatalakeFileSystem(AbstractFileSystem):
         )
 
         for file in files:
-            if "type" in file and file["type"] == "DIRECTORY":
-                file["type"] = "directory"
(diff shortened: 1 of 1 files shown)


hayesgb:
Can you pull the most recent changes from master into this branch and perform another push?  Seems there was an update to ADO which broke the linting step for CI.  It's been fixed.

dfigus:
@hayesgb I rebased onto master, but the build fails. However it looks unrelated to my changes.

PR #116, opened 2020-09-29
Get file

hayesgb:
Converts get_file method to async

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1154,7 +1154,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
             await self.put_file(lpath, rpath, delimiter)
 
     def put_file(self, lpath, rpath, delimiter="/", **kwargs):
-        """ Synchronous call to async version """
+        """ Alias synchronous call to async version """
         self.put(lpath, rpath, recursive=False, delimiter="/")
 
     def put(self, lpath, rpath, recursive=False, **kwargs):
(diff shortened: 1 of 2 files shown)


adlfs/spec.py
```diff
@@ -1169,7 +1174,7 @@ def put(self, lpath, rpath, recursive=False, **kwargs):
         rpaths = other_paths(lpaths, rpath)
 
         for lpath, rpath in zip(lpaths, rpaths):
-            maybe_sync(self.put_file, self, lpath, rpath, **kwargs)
+            maybe_sync(self._put_file, self, lpath, rpath, **kwargs)
```
martindurant, line 1177: The superclass method does this well no?
If not, recommend you still make an `async def _put`, and in there use `asyncio.gather`, so that you proprecc concurrently and get the possible speedup.

adlfs/spec.py
```diff
@@ -1152,6 +1153,10 @@ async def put_file(self, lpath, rpath, delimiter="/", **kwargws):
             await cc.create_container()
             await self.put_file(lpath, rpath, delimiter)
 
+    def put_file(self, lpath, rpath, delimiter="/", **kwargs):
```
martindurant, line 1156: This should be done automatically by the machinery in fsspec.asyn

martindurant: Plus, there is a decorator `async_wrapper` you could use

PR #132, opened 2020-11-16
Utilize async credentials for async blob storage access.

### LINKED ISSUE #131: AzureBlobFileSystem fails to open file with service principle authentication
Using a connection string for a storage account, this code is successful:

```python
fs = AzureBlobFileSystem(account_name='xxxxx', connection_string='xxxxxx')
with fs.open('/container/blob.txt') as f:
  txt = f.read()
```

However, when using a service principle to authenticate (in which the SP credentials are known to work when directly using the Azure SDK), I get the following error:

```python
import os
 fs = AzureBlobFileSystem(account_name='xxxxx', tenant_id=os.environ['AZURE_TENANT_ID'], 
    client_id=os.environ['AZURE_CLIENT_ID'], client_secret=os.environ['AZURE_CLIENT_SECRET'])
with fs.open('/container/blob.txt') as f:
  txt = f.read()
```
I get the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/conda/lib/python3.6/site-packages/fsspec/spec.py", line 844, in open
    **kwargs
  File "/opt/conda/lib/python3.6/site-packages/adlfs/spec.py", line 1259, in _open
    **kwargs,
  File "/opt/conda/lib/python3.6/site-packages/adlfs/spec.py", line 1350, in __init__
    self.details = self.fs.info(self.path)
  File "/opt/conda/lib/python3.6/site-packages/adlfs/spec.py", line 492, in info
    return maybe_sync(self._info, self, path)
  File "/opt/conda/lib/python3.6/site-packages/fsspec/asyn.py", line 80, in maybe_sync
    return sync(loop, func, *args, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/fsspec/asyn.py", line 51, in sync
    raise exc.with_traceback(tb)
  File "/opt/conda/lib/python3.6/site-packages/fsspec/asyn.py", line 35, in f
    result[0] = await future
  File "/opt/conda/lib/python3.6/site-packages/adlfs/spec.py", line 509, in _info
    out = await self._ls(self._parent(path), **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/adlfs/spec.py", line 687, in _ls
    async for next_blob in blobs:
  File "/opt/conda/lib/python3.6/site-packages/azure/core/async_paging.py", line 154, in __anext__
    return await self.__anext__()
  File "/opt/conda/lib/python3.6/site-packages/azure/core/async_paging.py", line 157, in __anext__
    self._page = await self._page_iterator.__anext__()
  File "/opt/conda/lib/python3.6/site-packages/azure/core/async_paging.py", line 99, in __anext__
    self._response = await self._get_next(self.continuation_token)
  File "/opt/conda/lib/python3.6/site-packages/azure/storage/blob/aio/_list_blobs_helper.py", line 75, in _get_next_cb
    use_location=self.location_mode)
  File "/opt/conda/lib/python3.6/site-packages/azure/storage/blob/_generated/aio/operations_async/_container_operations_async.py", line 1324, in list_blob_hierarchy_segment
    pipeline_response = await self._client._pipeline.run(request, stream=False, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/azure/core/pipeline/_base_async.py", line 215, in run
    return await first_node.send(pipeline_request)
  File "/opt/conda/lib/python3.6/site-packages/azure/core/pipeline/_base_async.py", line 83, in send
    response = await self.next.send(request)  # type: ignore
  File "/opt/conda/lib/python3.6/site-packages/azure/core/pipeline/_base_async.py", line 83, in send
    response = await self.next.send(request)  # type: ignore
  File "/opt/conda/lib/python3.6/site-packages/azure/core/pipeline/_base_async.py", line 83, in send
    response = await self.next.send(request)  # type: ignore
  [Previous line repeated 3 more times]
  File "/opt/conda/lib/python3.6/site-packages/azure/core/pipeline/_base_async.py", line 81, in send
    await _await_result(self._policy.on_request, request)
  File "/opt/conda/lib/python3.6/site-packages/azure/core/pipeline/_tools_async.py", line 32, in await_result
    return await result  # type: ignore
  File "/opt/conda/lib/python3.6/site-packages/azure/core/pipeline/policies/_authentication_async.py", line 37, in on_request
    self._token = await self._credential.get_token(*self._scopes)  # type: ignore
TypeError: object AccessToken can't be used in 'await' expression
```

PR with a fix is incoming.

**Environment**:

- Dask version: N/A
- Python version: 3.8
- Operating System: `ubuntu:18.04` dockerhub image

lossyrob:
#131 s is caused by the need to use [async credentials](https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#async-credentials) in the Aysnc clients of AzureBlobFileSystem (shoutout to @TomAugspurger for finding the fix). 

This PR modifies the AzureBlobFileSystem utilize async credentials. This requires the creation of both async and synchronous credentials, as the AzureBlobFile utilizes synchronous clients and takes credentials from the parent filesystem.

I tested this against my own use case; however I didn't find a way to create a unit test for these changes as Azurelite doesn't support service principal authentication. If there is a way to test this that you'd like me to add, please let me know!

Fixes #131

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -374,7 +374,12 @@ class AzureBlobFileSystem(AsyncFileSystem):
             and self.sas_token is None
             and self.client_id is not None
         ):
-            self.credential = self._get_credential_from_service_principal()
+            (
+                self.credential,
+                self.sync_credential,
+            ) = self._get_credential_from_service_principal()
+        else:
+            self.sync_credential = None
         self.do_connect()
 
     @classmethod
@@ -412,16 +417,26 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
         Returns
         -------
-        Credential
+        Tuple of (Async Credential, Sync Credential).
         """
         from azure.identity import ClientSecretCredential
+        from azure.identity.aio import (
+            ClientSecretCredential as AIOClientSecretCredential,
+        )
 
-        sp_token = ClientSecretCredential(
+        async_credential = AIOClientSecretCredential(
             tenant_id=self.tenant_id,
             client_id=self.client_id,
             client_secret=self.client_secret,
         )
-        return sp_token
+
+        sync_credential = ClientSecretCredential(
+            tenant_id=self.tenant_id,
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+        )
+
+        return (async_credential, sync_credential)
 
     def do_connect(self):
         """Connect to the BlobServiceClient, using user-specified connection details.
@@ -1384,9 +1399,9 @@ class AzureBlobFile(io.IOBase):
             self.fs.account_url: str = (
                 f"https://{self.fs.account_name}.blob.core.windows.net"
             )
-            if self.fs.credential is not None:
+            if self.fs.sync_credential is not None:
                 self.container_client = BlobServiceClient(
-                    account_url=self.fs.account_url, credential=self.fs.credential
+                    account_url=self.fs.account_url, credential=self.fs.sync_credential
                 ).get_container_client(self.container_name)
             elif self.fs.connection_string is not None:
                 self.container_client = BlobServiceClient.from_connection_string(


adlfs/spec.py
```diff
@@ -374,7 +374,12 @@ def __init__(
             and self.sas_token is None
             and self.client_id is not None
         ):
-            self.credential = self._get_credential_from_service_principal()
+            (
```
TomAugspurger, line 377: Maybe causing a linting issue: https://dev.azure.com/dask-dev/dask/_build/results?buildId=1724&view=logs&j=d090b295-8b65-5b33-df45-4b0e4ec519f8&t=34b50283-3bbf-53e1-90cc-f6f26ef783f5. Can you run `black` on your changes (or set up pre-commit to do it automatically).

TomAugspurger (commented):
Looks good to me at a glance (other than the linting issue).

Do you have a sense for whether `AzureBlobFile` can be made to work with an async credential? It seems a bit strange to carry both around, and if we're forced to choose one then async is likely the way to go (@martindurant might have experience here from s3fs and gcsfs, though I don't know if they have async variants of auth).

adlfs/spec.py
lossyrob: Yep - just rebased with the linting fixes.

lossyrob:
Thanks for taking a look! Unclear why there's both async and sync - perhaps @hayesgb could say? I agree moving `AzureBlobFile` to async IO would be the move if possible; perhaps that can be recorded for a follow up issue if needed.

martindurant:
Neither s3fs nor gcsfs have async-specific auth, this is initially done at instantiation and blocks. It is possible for more than one reauth to be in progress at once, but this isn't a race to worry about, since each would return valid credentials.

TomAugspurger:
OK, thanks Martin.

@hayesgb any objections to merging and possibly consolidating the sync and async auth as a followup? I have write access here so I can merge, just don't want to step on toes since I haven't been active here for a while.

hayesgb:
Thanks guys.  @TomAugspurger if you want to merge it in, feel free.  If you don't have time, I'll do it, and add a release tomorrow.

And yes, the intent is to get the AzureBlobFile to async as well.  The challenge I encountered is mixing sync and async operations from fsspec and the azure sdk, and the performance gains on the filesystem operations made it worth making this an intermediate step.  Apprecaite the help.

TomAugspurger:
Gotcha, thanks. This is all very new, but I suspect that we'll have more and more operations moving over to async (in fsspec) will hopefully make being all-async for auth here easier.

Issue #135, opened 2020-11-20
fs.find() returns a list instead of a dictionary


ldacey:
**What happened**:
I am running into errors with ds.dataset() from pyarrow which expects a dictionary instead of a list from fs.find() from adlfs. 

**What you expected to happen**:
The code seems to show that a dictionary should be returned if detail=True, but I wasn't able to get this to work.

**Minimal Complete Verifiable Example**:
dev is my Azure Blob container and testing is a dataset written with pq.write_to_dataset() with two partition columns being used.

```python
fs = fsspec.filesystem(protocol="abfs", account_name=name, account_key=key)

print(type(fs.find("dev/testing", detail=True)))
<class 'list'>

print(type(fs.find("dev/testing")))
<class 'list'>

fs = adlfs.AzureBlobFileSystem(account_name=name, account_key=key)
print(type(fs.find("dev/testing", detail=True)))
<class 'list'>
```

**Anything else we need to know?**:

**Environment**:
I installed this on a fresh PC using conda forge. Python was not installed previously, so I do not think there should be any conflicts with versions.

- Dask version: 2.23 typically but I do **not** have it installed in this example
- Python version: 3.8.6
- Operating System: Ubuntu 20.04
- Install method (conda, pip, source): conda env file

```
name: pyarrow
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.8
  - pandas
  - numpy
  - pyarrow
  - adlfs
  - azure-storage-blob
```

Full conda list output:

```
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
abseil-cpp                20200225.2           he1b5a44_2    conda-forge
adal                      1.2.5              pyh9f0ad1d_0    conda-forge
adlfs                     0.5.7              pyhd8ed1ab_0    conda-forge
aiohttp                   3.7.3            py38h25fe258_0    conda-forge
arrow-cpp                 2.0.0           py38ha9d6ab6_6_cpu    conda-forge
async-timeout             3.0.1                   py_1000    conda-forge
attrs                     20.3.0             pyhd3deb0d_0    conda-forge
aws-c-common              0.4.59               h36c2ea0_1    conda-forge
aws-c-event-stream        0.1.6                had2084c_6    conda-forge
aws-checksums             0.1.10               h4e93380_0    conda-forge
aws-sdk-cpp               1.8.70               h57dc084_1    conda-forge
[...]

PR #136, opened 2020-11-21
Updated mkdir and find to pass kwargs to async implementations

hayesgb:
Fix for #135


hayesgb:
@ldacey - Not sure if you saw this discussion at fsspec.  Link to [here](https://github.com/intake/filesystem_spec/issues/562) for any input you may have.

PR #141, opened 2020-11-30
Add metada support to the ls / find / info commands

### LINKED ISSUE #142: Add metadata support to AzureBlobFileSystem
The `AzureBlobFileSystem` doesn't expose the metadata of blobs to the caller.

Some other properties are also not exposed such as: 
`creation_time`, `last_modified`, `last_accessed_on`, `deleted_time`, `deleted`.

Currently there is no way to retrieve the metadata of a blob using the info() command. 

What about having a whitelist of properties that should be forwarded to the caller ?
 
Here is a first draft of a list of what might be useful for some use cases
- `metadata`
- `creation_time`
- `deleted`
- `deleted_time`
- `last_modified`
- `content_time`
- `content_settings`
- `remaining_retention_days`
- `archive_status`
- `last_accessed_on`
- `tags`
- `tag_count`

Darune:
Also add the following blob properties when detail is set to True
- creation_time
- deleted
- deleted_time
- last_modified
- content_time
- content_settings
- remaining_retention_days
- archive_status
- last_accessed_on
- tags
- tag_count

Fixes #142

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -29,6 +29,21 @@ from fsspec.utils import infer_storage_options, tokenize
 
 logger = logging.getLogger(__name__)
 
+FORWARDED_BLOB_PROPERTIES = [
+    "metadata",
+    "creation_time",
+    "deleted",
+    "deleted_time",
+    "last_modified",
+    "content_time",
+    "content_settings",
+    "remaining_retention_days",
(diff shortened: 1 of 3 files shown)


hayesgb:
@Darune -- Thanks!

PR #148, opened 2020-12-09
Add finalizer to AzureBlobFileSystem to avoid having unclosed service

### LINKED ISSUE #129: Error:asyncio:Unclosed connector
**What happened**:
When running upload/download scripts using adlfs, we get a warning on every run:

```
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f2012e55820>
ERROR:asyncio:Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7f2012ebcfa0>, 10699.800393553)]']
connector: <aiohttp.connector.TCPConnector object at 0x7f2012e5585
```

While it is just a warning, it can be scary for end-users as it seems that something went wrong. 

No warning should be raised

```python
import fsspec

fs: AzureBlobFileSystem = fsspec.filesystem("az", account_name="andersdatalake",
                                          account_key="my_account_key")

with open("hello_fsspec.txt", mode="w") as f:
    f.write("Hello from fsspec!\n")

fs.put("hello_fsspec.txt", "/raw/hello_fsspec.txt")
```

My immediate guess is that it comes from not closing the credentials transport session per this issue: Azure/azure-sdk-for-python/issues/8990 and this PR: Azure/azure-sdk-for-python/pull/9090

If that is the source, I imagine the fix would be to call `close` on the credentials object, but I'm not sure where in the lifecycle to do that.

**Environment**:

- adlfs: 0.5.5
- Dask version: N/A
- Python version: 3.8
- Operating System:Ubuntu 20.04

Darune:
This should fix the Error print on AzureBlobFileSystem garbage collection

Fixes #129

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -8,6 +8,7 @@ from glob import has_magic
 import logging
 import os
 import warnings
+import weakref
 
 from azure.core.exceptions import ResourceNotFoundError, ResourceExistsError
 from azure.storage.blob._shared.base_client import create_configuration
@@ -397,6 +398,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         else:
             self.sync_credential = None
         self.do_connect()
+        weakref.finalize(self, maybe_sync, self.service_client.close, self)
 
     @classmethod
     def _strip_protocol(cls, path: str):


hayesgb:
Thanks @Darune !

PR #153, opened 2020-12-14
Change "find()" of AzureBlobFileSystem to match gcsfs implementation

### LINKED ISSUE #152: Harmonize AzureBlobFilesystem `find` with gcsfs one
As discussed in a corresponding gcsfs [issue](https://github.com/dask/gcsfs/issues/317)

It would be great if the find command of both (and maybe all) cloud storage behave the same. 

mostly find should do only prefix match on blob names and avoid recursion
this means removing the `maxdepth` argument in favor of:
- a path (a trailing `/` will be appended to this path to ensure this is a path
- a prefix (a partial matching string that will be concatenated to the path)

in simple the find method will befhave like the following regular glob on a container:
`^{path.rstrip(`/`)}/{prefix}*`

but this matching will be done server side allowing to get all the matching blobs in one API call.

Darune:
Previously the find would take a path and recursively list all the
sub folders until maxdepth(default to None) is reached, this was mostly
used by glob().

This changes the behavior to the following one:

Take a path (a directory) and a prefix (partial match)
and list ervery blob that matches "^{path}/{prefix}*"
This is done in one single API call, no recursions.
The `maxdepth` argument of "find" has been dropped since it doesn't have
any meaning now all the matching blobs will be returned.

It suffers the same 'bug' regarding withdirs kwargs:

  The following structure might bring wrong dir results:
    - directory_a/a/b/c/file.txt
  when querying path: `directory_a` prefix: `a` and withsubdirs
  only directory_a/b/c will be returned as a directory which means
  that only folder with a file will be returned, Reading the from
  gcsfs, it looks like this is normal behavior.
  It might not be considered a bug thou

The glob function behavior however remains unchanged since this is the
only way we can do proper complex pattern matching with branch elimination
which might, for some, be useful.

https://github.com/dask/adlfs/commit/b490a8cbc637dc6547c69fb29c8abb5a414a536a introduces a breaking change let me know if you have any annotation / process when the api breaks

Fixes #152

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -601,7 +601,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
             root = ""
             depth = 20 if "**" in path else 1
 
-        allpaths = await self._find(
+        allpaths = await self._glob_find(
             root, maxdepth=depth, withdirs=True, detail=True, **kwargs
         )
         pattern = (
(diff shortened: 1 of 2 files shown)


hayesgb:
Appreciate the contribution!!

PR #157, opened 2020-12-17
PERF: Improve performance of exists

TomAugspurger:
Closes https://github.com/dask/adlfs/issues/156.

There is a slight behavior change here. Since we aren't going through
`ls`, we don't cache that this file exists anymore. I *think* that's
probably OK, but we may be able to cache it (assuming the BlobClient
has all the information we need).


TomAugspurger:
Merging later today if there aren't any objections.

hayesgb:
Looks great to me!

TomAugspurger:
Thanks!

PR #159, opened 2020-12-19
Add python requires

TomAugspurger:


setup.py
```diff
@@ -9,13 +9,23 @@
...
+        "Programming Language :: Python :: 3.9",
+    ],
     description="Access Azure Datalake Gen1 with fsspec and dask",
     url="https://github.com/hayesgb/adlfs/",
     maintainer="Greg Hayes",
     maintainer_email="hayesgb@gmail.com",
     license="BSD",
     keywords=["file-system", "dask", "azure"],
     packages=["adlfs"],
+    python_requires=">3.6",
```
TomAugspurger, line 28: This matches fsspec's current setting. Probably best to just track that.

raybellwaves:
once this merges can add to the host (and run) here https://github.com/conda-forge/adlfs-feedstock/blob/master/recipe/meta.yaml#L18. Can add you the maintainers there as well.

raybellwaves:
bumping on conda as well https://github.com/conda-forge/adlfs-feedstock/pull/24

PR #164, opened 2021-01-12
Speedup

hayesgb:
Implements asynchronous writes and direct async upload within pipe_file


adlfs/spec.py
```diff
@@ -1473,6 +1484,7 @@ def __init__(
         self.end = None
         self.start = None
         self.closed = False
+        self.loop = self.fs.loop or get_loop()
```
TomAugspurger, line 1487: Is this used anywhere?

TomAugspurger (commented):
Do we have a policy on when (aysnc) `azure.storage.blob` clients are closed?

IIUC, when we create the AzureBlobFile we set `self.container_client` to an open (I think) client. And then we might use it  and close it with

```python
async with self.container_client as cc:
    cc.<foo>
```

(at least I assume that `__exit__` will close the client. Is that necessary, or should we only use the context manager for BlobServiceClients, since we'll presumably be making many of them?

adlfs/spec.py
hayesgb: self.loop is expected by sync_wrapper.

TomAugspurger:
These changes look nice. It'd be nice to deliberately decide on a policy for when we open / close a connection, but that shouldn't be a blocker here.

adlfs/spec.py
hayesgb: Thanks for the feedback on the context manager for the container client @TomAugspurger.  Agree that having this as a policy will help with clarity around the managing the asynchronous clients.

jhamman:
Thanks for this @hayesgb!

PR #173, opened 2021-02-08
fix memory leak azure file

btel:
adfls causes unbounded memory allocation when adding mutliple files (or same file many times).

Here is some code to reproduce it:

```
import adlfs

data = b"0123456"
bucket_name = "testleak"

account_name = "..."
account_access_key = "..."
conn_str = "..."

fs = adlfs.AzureBlobFileSystem(account_name=account_name, connection_string=conn_str)

while True:
    with fs.open(bucket_name + "/test_file.bin", 'wb') as f:
        f.write(data)
```

which produces this linear allocated memory size:

simple_adlfs

This is related to weakref.finalize call that will keep the references of the AzureBlobFile object. According to python docs:

> Note: It is important to ensure that func, args and kwargs do not own any references to obj, either directly or indirectly, since otherwise obj will never be garbage collected. In particular, func should not be a bound method of obj. 

https://docs.python.org/3/library/weakref.html#weakref.finalize

The close function can be called explicitly or in a context manager as shown in tests.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1519,7 +1519,6 @@ class AzureBlobFile(AbstractBufferedFile):
             cache_options["trim"] = kwargs.pop("trim")
         self.metadata = None
         self.kwargs = kwargs
-        weakref.finalize(self, maybe_sync, self.container_client.close, self)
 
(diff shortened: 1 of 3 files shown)


hayesgb:
Thanks!

PR #181, opened 2021-02-15
get presigned url for blob with a timeout

madhur-tandon:


madhur-tandon:
It seems like the tests fail when all 3 of the fields i.e.
`account_name`, `connection_string` and `account_key` are supplied while instantiating
`fs = AzureBlobFileSystem(...)` object

This is weird since we should be able to supply all of them.

For this feature to work, we need to supply `account_key` as well.

I have tested the implementation outside the testing framework and there is no issue with it.
I could use some help regarding what I am supposed to do?

hayesgb:
When you said to works outside the testing framework, do you mean this PR works when used against a live Azure account, or do you mean it works when using azure-storage-blob against Azurite?

madhur-tandon:
I meant that this PR works when used against a live Azure Account.

hayesgb:
Do either of these describe the issue?  https://stackoverflow.com/questions/50874498/content-disposition-header-ignored-on-download-with-azure-blob-storage

Or

https://github.com/Azure/Azurite/issues/470

madhur-tandon:
I am not sure, but it's definitely related to the authorization header. From the logs, the message is this:

```
azure.core.exceptions.ClientAuthenticationError: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.

AuthenticationErrorDetail:The MAC signature found in the HTTP request 'GexAw1uNZuQvdFoJOILwBrLX/epRnWirjVOVvJ5lSPY=' is not the same as any computed signature. Server used following string to sign: 'GET'
```

It's definitely not the 2nd issue you listed because I am not using the `content_disposition` argument of the `generate_blob_sas` function.

Could you please investigate / point me to some next steps?

wolfv:
Ping on this! :)

adlfs/spec.py
```diff
@@ -1269,6 +1272,32 @@ def cat(self, path, recursive=False, on_error="raise", **kwargs):
...
+        sas_token = generate_blob_sas(
+            account_name=self.account_name,
+            container_name=container_name,
+            blob_name=blob,
+            account_key=self.account_key,
+            permission=BlobSasPermissions(read=True),
+            expiry=datetime.utcnow() + timedelta(seconds=expires),
+        )
+
+        cc = self.service_client.get_container_client(container_name)
```
hayesgb, line 1296: Please convert this to async, as:
```
async with self.service_client.get_blob_client(container_name, blob) as bc:
    url = BlobClient.from_blob_url(bc.url, credential=XXX).url

return url
```

This will help avoid hanging connections.  Would also prefer to see these imports moved inside the method call.

adlfs/tests/test_spec.py
```diff
@@ -1137,6 +1137,22 @@ def test_cat(storage):
     fs.rm("catdir/catfile.txt")
 
 
+def test_url(storage):
```
hayesgb, line 1140: Based on [this](https://stackoverflow.com/questions/50874498/content-disposition-header-ignored-on-download-with-azure-blob-storage), it doesn't appear the tests will work with Azurite.

madhur-tandon: Should I skip the tests altogether then? aka remove them?
Then we would only have the functionality without the tests.
Let me know, Thanks!

adlfs/spec.py
hayesgb: I was thinking about this a bit more.  I'm not sure you need the `BlobClient.from_blob_url(xyz...).url`

Shouldn't the following work?
```
async with self.service_client.get_blob_client(container_name, blob) as bc:
    url = bc.from_blob_url(bc.url, credential=XXX).url
return url
```

adlfs/tests/test_spec.py
hayesgb: It would be best if you can create a mock for fs, rather than using Azurite.

adlfs/spec.py
madhur-tandon: I have incorporated this change.

adlfs/tests/test_spec.py
madhur-tandon: Seems like I cannot even use just `fs.mkdir("catdir")` aka I removed everything in the test apart from `fs.mkdir` [even calls to `fs.url(...)`]. But even a simple operation such as `fs.mkdir` doesn't work. Like I posted in my first comment, the issue comes when I supply all 3 arguments i.e.

```
fs = AzureBlobFileSystem(
        account_name=ACCOUNT_NAME, account_key=KEY, connection_string=CONN_STR
    )
```
instead of supplying only 2 of them as done in the other tests.

Is this still fixable by mocking?
Also, could you give me some pointers about mocking?
I don't know / haven't tried `mocking` before.

hayesgb: Found the issue.  The following fixes it in the unit tests.  You should confirm it works as expected...

```
        async with self.service_client.get_blob_client(container_name, blob) as bc:
            url = f"{bc.url}?{sas_token}"
        return url
```

hayesgb: Calling .url on the end was stripping a "/" between the http address and the account name

madhur-tandon: While this doesn't change the behaviour with a live account (which is good), it is still unable to fix the tests for me. I think the reason for tests failing is not because of using `.url`. It's probably because we use all 3 fields i.e. `account_name=ACCOUNT_NAME, account_key=KEY, connection_string=CONN_STR` while instantiating the file system object. Regardless, I have pushed a commit with your suggested change. Let's see if the tests pass on the CI.

madhur-tandon: Seems like the tests passed on the CI. Although they still fail for me locally which is strange. Regardless, this PR should be ready.

hayesgb:
Thanks!

Issue #183, opened 2021-02-15
Is account_name required when connecting through connection_string


isidentical:
It is only used in `do_connect` but not when connecting through a `connection_string`, would it make sense to remove that restriction or make it if connection_string is not provided;
https://github.com/dask/adlfs/blob/2bd9d9302e5d3511ce636246cbd2d7770000f00c/adlfs/spec.py#L486-L489
https://github.com/dask/adlfs/blob/2bd9d9302e5d3511ce636246cbd2d7770000f00c/adlfs/spec.py#L375-L382

hayesgb:
Take a look [here](https://github.com/dask/adlfs/tree/num_183) and let me know what you think.

isidentical:
It seems perfect (maybe use `os.getenv` instead here, since that key might not be present);
https://github.com/dask/adlfs/commit/d66f2da7f2d71a2effabb0eef9aaa71d01aa0e7e#diff-c36c47c3116b1b34239a3521711bee4796cde28afb220a3a8366f6caa546302dR375

hayesgb:
Good catch.  Thanks.

PR #184, opened 2021-02-15
Num 183

hayesgb:


hayesgb:
Addresses #183

isidentical:
@hayesgb would it be possible for you to make a release including this fix, it would be much appreciated!

PR #187, opened 2021-02-18
When connecting preserve a reference for the underlying issue

isidentical:
For people who might want to print a pretty message according to the underlying cause (like a network error, authorization error etc) it is useful to keep a reference (in `__cause__`).


TomAugspurger:
Thanks!

PR #191, opened 2021-03-03
Proper folder handling

isidentical:
Here is a simple demonstration of what I've been trying to solve;
```py
import secrets

BUCKET = 'test-8/'
FOLDER = f'test-8/{secrets.token_hex(12)}/'

fs = AzureBlobFileSystem(account_name=_NAME, account_key=_KEY)
fs.touch(FOLDER + 'foo')
fs.touch(FOLDER + 'bar')
fs.invalidate_cache()

fs = AzureBlobFileSystem(account_name=_NAME, account_key=_KEY)
print(fs.exists(FOLDER))
print(FOLDER in fs.ls(fs._parent(FOLDER)))
```
The `exists()` calls results with False even though it is a folder (can be shown by the next call). There are some places in the code which should clear trailing slash when passing (to use with `_ls_from_cache`) and also `.exists()` need to fall back to `.info()` in the case of cache being empty and the searched thing is not a file but rather a directory. So when fall backed into `super().exists()` which wraps `.info()` under the hood which searches for this directory.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1216,6 +1216,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         except FileNotFoundError:
             pass
 
+        full_path = path
         container_name, path = self.split_path(path)
 
         if not path:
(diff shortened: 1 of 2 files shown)


isidentical:
Actually awaiting https://github.com/intake/filesystem_spec/issues/554

hayesgb:
Closing this as several commits behind master.

PR #206, opened 2021-04-07
fsspec 0.9 migration

### LINKED ISSUE #205: Adjust for fsspec asyn changes and remove fsspec pin <=0.8.7
Based on [#581](https://github.com/intake/filesystem_spec/pull/581) on fsspec, there will be a significant change to the fsspec async API.  Until that work is finalized, adlfs will be pinned to fsspec <=0.8.7.

Pin to be removed when the work to fsspec is complete, and adlfs updated to align with changes.

isidentical:
Resolves #205

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ azure-storage-blob>=12.5.0,<12.7.0
 azure-identity
 dask[dataframe]
 docker>=4.1.0,<5.0.0
-fsspec>=0.8.2,<=0.8.7
+fsspec>=0.9
 msrestazure
 pyarrow
 pytest>=5.0,<6.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ azure-storage-blob>=12.5.0,<12.7.0
 azure-identity
 dask[dataframe]
 docker>=4.1.0,<5.0.0
-fsspec>=0.9,<0.10.0
+fsspec>=0.9
 msrestazure
 pyarrow
 pytest>=5.0,<6.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,7 +5,7 @@ azure-storage-blob>=12.5.0,<12.7.0
 azure-identity
 dask[dataframe]
 docker>=4.1.0,<5.0.0
-fsspec>=0.9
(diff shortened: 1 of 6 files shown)


isidentical:
@hayesgb apparently there were 2 cases (one in `_rm_file`, and the other one in `_get_file`) that required a `.rstrip(delimiter)` though I couldn't find why they popped up now. Other than that, if I am not missing anything there is no async code that calls another synchronized async function so I just migrated maybe_sync's into sync

dev-requirements.txt
```diff
@@ -5,7 +5,7 @@ azure-storage-blob>=12.5.0,<12.7.0
 azure-identity
 dask[dataframe]
 docker>=4.1.0,<5.0.0
-fsspec>=0.8.2,<=0.8.7
+fsspec>=0.9,<0.10.0
```
TomAugspurger, line 8: fsspec doesn't use semantic versioning, so the `<0.10` doesn't add any benefit.

```suggestion
fsspec>=0.9
```

requirements.txt
```diff
@@ -3,6 +3,6 @@ azure-core>=1.5.0
 azure-datalake-store>=0.0.46,<0.1
 azure-identity
 azure-storage-blob>=12.5.0,<12.7.0
-fsspec>=0.8.2,<=0.8.7
+fsspec>=0.9.0,<0.10.0
```
TomAugspurger, line 6:
```suggestion
fsspec>=0.9.0
```

setup.py
```diff
@@ -33,7 +33,7 @@
         "azure-datalake-store>=0.0.46,<0.1",
         "azure-identity",
         "azure-storage-blob>=12.5.0,<12.7.0",
-        "fsspec>=0.8.2,<=0.8.7",
+        "fsspec>=0.9.0,<0.10.0",
```
TomAugspurger, line 36:
```suggestion
        "fsspec>=0.9.0",
```

TomAugspurger (commented):
Thanks for working on this.

isidentical:
@hayesgb it would be amazing to have a new release (so that we can install newer versions of adlfs with other types of remotes [eg gcsfs] at the same time)

hayesgb:
Done.

isidentical:
Thanks

Issue #201, opened 2021-03-23
Try DefaultAzureCredential by default?


dylanfprice:
Use case: accessing blob storage from an Azure VM using a managed identity.

Using [fsspec](https://filesystem-spec.readthedocs.io/en/latest), I was able to get it to work via:
```python
from azure.identity.aio import DefaultAzureCredential         
import fsspec

fsspec.filesystem('az', account_name='myaccount', credential=DefaultAzureCredential())
```

However, using [petl](https://petl.readthedocs.io/en/stable/io.html#petl.io.remotes.RemoteSource), which uses fsspec under the hood, I am not sure how to inject the `DefaultAzureCredential`.

It might be nice to try the `DefaultAzureCredential` by default, since it is meant to cover a range of common use cases.

dylanfprice:
Worked around this via the following. Not perfect but handles my use cases.
```python
import os

import adlfs
from azure.identity.aio import DefaultAzureCredential


class AzureBlobFileSystemWithDefaultCredentials(adlfs.AzureBlobFileSystem):
    def __init__(self, *args, **kwargs):
        has_credential = (
            "AZURE_STORAGE_ACCOUNT_KEY" in os.environ or "credential" in kwargs
        )
        if not has_credential:
            kwargs["credential"] = DefaultAzureCredential()
        super().__init__(*args, **kwargs)

# then before I try to open files
import fsspec
fsspec.register_implementation("az", AzureBlobFileSystemWithDefaultCredentials)
```

isidentical:
In dvc, we also have a special condition to use `DefaultAzureCredential()`. It would be awesome to fallback on this by default on adlfs itself too! 
https://github.com/iterative/dvc/blob/547db15d343515d7056c7af4c311ae41b139916e/dvc/fs/azure.py#L72-L73

thec0dewriter:
Hello @isidentical , @dylanfprice 
I tried to put together a PR, for falling back to use the DefaultAzureCredential. 
Let me know if this was kind of what were you expecting

dylanfprice:
Yeah I am not familiar with the adlfs codebase but that looks about right to me!

TomAugspurger:
This makes sense to me. I think the only potential issue to consider is that anonymous access to public storage accounts / blobs still works.

Anyone interested in submitting a pull request?

dylanfprice:
There is one started in #222

PR #222, opened 2021-04-17
Fallback for azure identity default authentication #201

thec0dewriter:
Methods implementation and readme update. #201 
The main goal of this would be to support Azure Managed Identities for credential free authentication.
This could be really useful for local development too as it allows to  pull Azure CLI and VSCode tokens

This is my first PR and I would be happy to get feedbacks.
Thanks!

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -346,6 +346,11 @@ class AzureBlobFileSystem(AsyncFileSystem):
         ...    client_id=CLIENT_ID, client_secret=CLIENT_SECRET)
     >>> abfs.ls('')
 
+    Authentication with environment based identity which can be  EnvironmentCredential, ManagedIdentityCredential,
+    SharedTokenCache(Windows only), AzureCLI identity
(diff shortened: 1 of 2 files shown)


thec0dewriter:
Hello @hayesgb,
I'm not sure who is the maintener but you seem to be the most active contributors. Could you please have a look at this. I would be happy to fix if  it needs something additional
Thank you!

adlfs/spec.py
```diff
@@ -400,12 +405,17 @@ def __init__(
...
-                self.credential,
-                self.sync_credential,
-            ) = self._get_credential_from_service_principal()
+            if self.client_id is not None:
+                (
+                    self.credential,
+                    self.sync_credential,
+                ) = self._get_credential_from_service_principal()
+            else:
+                (
```
hayesgb, line 415: In this case, if a defaultcredential is returned, then line #513 will have a value for creds, which may attempt to authenticate against a public container.  Can you add an "anon" parameter to handle access to a public container?

PR #228, opened 2021-05-06
Haneld the situation that Azure Datalake Gen1 returns filename

### LINKED ISSUE #227: AzureDatalakeFileSystem.ls should distinct the result on it's type dict or str
**What happened**:
AzureDatalakeFileSystem.ls(rpath, detail=True) crash when the results path contains "type" or "length" as part of path

No crash when "type" or "length" is part of path.

```python
# Prepare Azure Datalake Gen1 with a folder contains "type" as part of path, e.g. data/_type_/a
fs = AzureDatalakeFileSystem(...)
fs.ls("data/_type_/a", detail=False)  # detail=False is defailt, just for emphasize

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-22-c9e51a630e9b> in <module>
----> 1 adl1.ls(rpath1+"/dag_type", detail=False)# +"/markdown")

/mdo/adlfs/spec.py in ls(self, path, detail, invalidate_cache, **kwargs)
    134         for file in files:
    135             if "type" in file:
--> 136                 file["type"] = file["type"].lower()
    137             if "length" in file:
    138                 file["size"] = file["length"]

TypeError: string indices must be integers
```

The return type of ls can be List[str] or List[Dict], it depends on detail=False or True.
When return type is List[Dict] (detail=True), the corrent code is fine.
When return type is List[str], and one of the path contains e.g. "type", the code at line 135 is actually check if "type" is a substring of the path. And if it returns true, line 136 crash.

**Environment**:

- Dask version: N/A
- Python version: 3.8
- Operating System: OSX / Ubuntu
- Install method (conda, pip, source): pip / conda

scott-zhou:
contains special sub-string

AzureDatalakeFileSystem.ls has special mechanism for type and length.
But the above special mechanism should only applicable when call ls
with detail=True and ls returns a list of dictionary.

When call ls with detail=False, it returns a list of str. Keyword in
for str has the meaning of check if one str is substring of another.
That is not what the mechanism should acting on.

Closes #227

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -140,7 +140,7 @@ class AzureDatalakeFileSystem(AbstractFileSystem):
             path=path, detail=detail, invalidate_cache=invalidate_cache
         )
 
-        for file in files:
+        for file in (file for file in files if type(file) is dict):
             if "type" in file:
                 file["type"] = file["type"].lower()
             if "length" in file:


hayesgb:
Thanks!

PR #229, opened 2021-05-10
Fix loop

hayesgb:
Adds ability to catch event loop when requesting async token through azure identity sdk

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -382,8 +382,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
             if k in kwargs
         }  # pass on to fsspec superclass
         super().__init__(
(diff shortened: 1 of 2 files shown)


isidentical:
@hayesgb just FYI, we recently experienced an issue regarding the combination of python3.8 + windows + proxies + aiohttp + azure. Unfortunately, the aiohttp bug is around for a couple of years, and there isn't much action taken regarding it (https://github.com/aio-libs/aiohttp/issues/4536). On DVC we currently host a piece of code that takes care of this issue (https://github.com/iterative/dvc/commit/fc4e31bbf8d4699ca71f11ea7479527f0a89336b). What do you think about also changing the policy on adlfs by default, since it might affect other users too?

PR #232, opened 2021-05-16
Fix bug for filter_blobs when target_path contains special character …

### LINKED ISSUE #231: ls on Azure Datalake Gen2 on a path has special character for regex returns not correct
**What happened**:
For Azure Datalake Gen2 (blob storage), when ls a path contains some special character for regex (e.g. a path contains +), it won't return correctly.

For example, if we have following files:
 data/root/e+f/file1.txt
 data/root/e+f/file2.txt
ls("data/root/e+f") will raise FileNotFoundError

It should able to handle path contains special character for regex, as soon as it it legal as a path.
For the above example, it should returns ["data/root/e+f/file1.txt", "data/root/e+f/file2.txt"]

The root cause is in utils.py, filter_blobs(...) function, it's using re module to check if the blob's name is a subpath of target_path, but it didn't handle the situation if target_path contains any special character for regex. 
```python
# Put your MCVE code here
import re
blobs = [
    {"name": "data/root/e+f/file1.txt"},
    {"name": "data/root/e+f/file2.txt"}
]
target_path = "data/root/e+f"
finalblobs = [
    b for b in blobs if re.search(r"\b" + target_path + r"(?=/)" + r"\b", b["name"])
]
print(finalblobs)  #in this case, no blob should be filtered, but actually, finalbolbs is an empty array
```

All the special character for regex and also legal as part of path need be take care if this function want to use regex to check subpath. 
Maybe use re module to check if a blob is subpath is a bit overhead. Use str.startswith might be much easier.

**Environment**:

- Dask version: N/A
- Python version: 3.7
- Operating System: Linux / OSX
- Install method (conda, pip, source): conda

scott-zhou:
Fix bug for filter_blobs when target_path contains special character for regex

Use str.startswith to check if a blob is come from target_path.
Maybe use regex is a bit overhead for this case.

Add the test case when the path has + as part of path, which also explains what is the situation.
The test case won't pass without this PR.

Closes #231

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -881,7 +881,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
             if len(output) == 1 and output[0]["type"] == "file":
                 # This handles the case where path is a file passed to ls
                 return output
-            output = await filter_blobs(output, target_path)
+            output = await filter_blobs(output, target_path, delimiter)
 
         return output
 
--- a/adlfs/tests/conftest.py
+++ b/adlfs/tests/conftest.py
@@ -53,6 +53,8 @@ def storage(host):
     container_client.upload_blob(
         "root/d/file_with_metadata.txt", data, metadata=metadata
     )
+    container_client.upload_blob("root/e+f/file1.txt", data)
+    container_client.upload_blob("root/e+f/file2.txt", data)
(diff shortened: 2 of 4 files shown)


hayesgb:
Thanks!

Issue #233, opened 2021-05-17
AzureDatalakeFileSystem need implement rm_file and rmdir methods


scott-zhou:
The remove file and remove directory functions are not implemented in AzureDatalakeFileSystem for Azure Datalake Gen1.
When call rm or rm_dir, it will ends up to AbstractFileSystem's implementation, which for rmdir is a empty function, and for rm_file(rm will also use rm_file), raise NotImplementedError.

PR #234, opened 2021-05-17
Implement rm_file and rmdir for AzureDatalakeFileSystem

scott-zhou:
Implement rmdir and rm_file for AzureDatalakeFileSystem

No need to implement rm method, it will use AbstractFileSystem's implementation and end up to rm_file.

#233

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -206,6 +206,14 @@ class AzureDatalakeFileSystem(AbstractFileSystem):
     def size(self, path):
         return self.info(path)["length"]
 
+    def rmdir(self, path):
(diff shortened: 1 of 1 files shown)


scott-zhou:
To be honest, I didn't implement any test case, which is not good. But the reason is the whole AzureDatalakeFileSystem doesn't have any unit test case, so it's a bit hard to do so for add these two small methods.

hayesgb:
Thanks @scott-zhou.  I'll merge this in.  Azure Blob Storage (Gen2) went GA shortly after this was created, and Gen1 has since been deprecated by Azure, so not much work has gone into that implementation.  Appreciate the PR!

PR #237, opened 2021-05-25
spec: support exists() on directories pre-cache

isidentical:
```
import secrets

BUCKET = 'bucket/'
FOLDER = f'{BUCKET}{secrets.token_hex(12)}/'

fs = AzureBlobFileSystem(**opts)
fs.touch(FOLDER + 'foo')
fs.touch(FOLDER + 'bar')
fs.invalidate_cache()

print(fs.exists(FOLDER))
fs.ls(fs._parent(FOLDER))
print(fs.exists(FOLDER))
```

While the first one is returning False, after the cache the second one is returning true. The reason is that, the `exists()` implementation first tries to check whether a such a directory exists by traversing the cache and when it fails it tries to check if a blob named `path` exists though if the directory is not cached, it doesn't check for that. This patch makes it fallback to the default implementation which is basically checking the result of `info()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 **Change Log**
+v0.7.5
+------
+- `exists()` calls now also checks whether a directory with that name exists or not. Previously this was only checked from the cache.
+
 v0.7.4
 ------
 - Added the location_mode parameter to AzureBlobFileSystem object, and set default to "primary" to enable Access Control Lists and RA-GRS access.  Valid values are "primary" and "secondary"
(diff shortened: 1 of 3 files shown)


TomAugspurger (approved):
Looks good. One small question.

isidentical:
> One small question.

I can't see the question.

TomAugspurger:
Strange, my review was lost. The gist was: should the `.rstrip` call be `.rstrip("/", 1)` to avoid cases where the path is `/path/to/file//` (if that's even allowed)?

TomAugspurger (approved):
CI is passing now. Planning to merge in a few hours.

TomAugspurger:
Thanks @isidentical!

PR #242, opened 2021-06-08
Find yield the exact match

### LINKED ISSUE #239: [Question] Any particular reason for not returning exact matches on find(..., prefix=...)
I was implementing the same API for the s3fs but the there is a slight edge case that I don't know whether done intentionally or not. When you are trying to `find('a/b/', prefix='file')` if there is a `a/b/file` in that same directory as well as `a/b/file_1` and `a/b/file_2` then it doesn't yield the exact match. 

https://github.com/dask/adlfs/blob/96958331854c5a67917d4c9a802a62720d091374/adlfs/spec.py#L934-L935

Is there any particular reason to excluding this match? If so, would it be possible to introduce some sort of keyword argument (like `include_full_match` or something like that) since otherwise we might need to do another API call to preserve the consistency of returns regardless of platform (s3fs/gcsfs/adlfs) which is not needed at all since the `_find()` method already finds that file if it exists.

isidentical:
Resolves #239

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -931,8 +931,6 @@ class AzureBlobFileSystem(AsyncFileSystem):
             infos = await self._details([b async for b in blobs])
             for info in infos:
                 name = info["name"]
-                if name == target_path:
-                    continue
                 parent_dir = self._parent(name).rstrip("/") + "/"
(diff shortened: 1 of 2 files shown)


TomAugspurger (approved):
CI is passing now. Planning to merge in a few hours.

PR #243, opened 2021-06-08
Implement `_cat_file`

### LINKED ISSUE #240: adlfs on fsspec upstream is broken due to cat_file
https://github.com/intake/filesystem_spec/pull/647/ added `_cat_file` as a dummy implementation (just raising a `NotImplementedError`) which now breaks the assumption of that the method will be inherited from the `AbstactFileSystem` so the CI is failing.

TomAugspurger:
Compatibility with the latest fsspec.

Closes #240.

cc @hayesgb, worth doing a release with this ASAP I think. I believe I have permissions, if you want me to do it.

It'd be nice to include https://github.com/dask/adlfs/pull/237 and https://github.com/dask/adlfs/pull/242 in the release too. We can merge those after this, since CI is failing without this.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1303,6 +1303,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
     pipe_file = sync_wrapper(_pipe_file)
 
     async def _cat_file(self, path, start=None, end=None, **kwargs):
+        path = self._strip_protocol(path)
         if end is not None:
             start = start or 0  # download_blob requires start if length is provided.
             length = end - start
@@ -1312,7 +1313,10 @@ class AzureBlobFileSystem(AsyncFileSystem):
         async with self.service_client.get_blob_client(
             container=container_name, blob=path
         ) as bc:
-            stream = await bc.download_blob(offset=start, length=length)
+            try:
+                stream = await bc.download_blob(offset=start, length=length)
+            except ResourceNotFoundError as e:
+                raise FileNotFoundError from e
             result = await stream.readall()
             return result
 
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1304,7 +1304,8 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
     async def _cat_file(self, path, start=None, end=None, **kwargs):
         if end is not None:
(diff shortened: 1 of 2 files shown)


adlfs/spec.py
```diff
@@ -1302,6 +1302,20 @@ async def _pipe_file(self, path, value, overwrite=True, **kwargs):
 
     pipe_file = sync_wrapper(_pipe_file)
 
+    async def _cat_file(self, path, start=None, end=None, **kwargs):
```
isidentical, line 1305: `strip_protocol` seems to be missing.

adlfs/spec.py
```diff
@@ -1302,6 +1302,20 @@ async def _pipe_file(self, path, value, overwrite=True, **kwargs):
...
+        if end is not None:
+            start = start or 0  # download_blob requires start if length is provided.
+            length = end - start
+        else:
+            length = None
+        container_name, path = self.split_path(path)
+        async with self.service_client.get_blob_client(
+            container=container_name, blob=path
+        ) as bc:
```
isidentical, line 1314: I guess it would be nice if this can be wrapped into `try/except` for catching `ResourceNotFoundError` though there are some more missing places so maybe we can add some sort of decorator do that automatically on all functions, ([an example](https://github.com/iterative/sshfs/blob/eddccf69abe6e867a451df57dff707412c3c6a4b/sshfs.py#L32-L52]) What do you think?

isidentical (approved):
looks good to me

TomAugspurger:
Thanks, those should be fixed.

adlfs/spec.py
TomAugspurger: Hmmm I could go either way on that. I'm a bit leery of decorators like this, since their behavior can be hard to understand since they're separated from the implementation of the method. But it's difficult to ensure that we properly handle this everywhere.

TomAugspurger:
I'll merge this, and then take a look at the other PRs.

Will hold off a bit on a release until @hayesgb has a chance to give a +1, but it'd be nice to do one today or tomorrow.

TomAugspurger:
0.7.6 is up at https://pypi.org/project/adlfs/0.7.6/ now.

PR #244, opened 2021-06-08
Handle use_listings_cache=False in ls

### LINKED ISSUE #230: `AzureBlobFileSystem.ls` implementation is incompatible with `use_listings_cache=False`
Currently `_ls` is implemented by writing things directly to `self.dircache` and then returning a result from them. See https://github.com/dask/adlfs/blob/faf4da0d5ef2f2ac6fcc8804ba433d042a27271d/adlfs/spec.py#L810-L812.

With `use_listings_cache=False`, this doesn't work since the dircache will always be empty.

I think this test captures the behavior.

```python
def test_nodircache(storage):
    fs = AzureBlobFileSystem(
        account_name=storage.account_name,
        connection_string=CONN_STR,
        use_listings_cache=False,
        skip_instance_cache=True,
    )
    result = fs.cat_file("data/top_file.txt")
    assert result
```

TomAugspurger:
Closes #230

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -737,7 +737,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         return_glob: bool
 
         """
-        logger.debug(f"abfs.ls() is searching for {path}")
+        logger.debug("abfs.ls() is searching for %s", path)
         target_path = path.strip("/")
         container, path = self.split_path(path)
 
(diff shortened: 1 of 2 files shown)


adlfs/spec.py
```diff
@@ -737,26 +737,31 @@ async def _ls(
         return_glob: bool
 
         """
-        logger.debug(f"abfs.ls() is searching for {path}")
+        logger.debug("abfs.ls() is searching for %s", path)
```
TomAugspurger, line 740: FYI, it's depending on how expensive string formatting is, it's better to use `"%"`-style format strings. That avoids calling `str(path)` if the level is higher than debug.

Issue #57, opened 2020-05-20
azure-storage-blob >=v12 causes slow, high memory dd.read_csv


mgsnuno:
`dd.read_csv(path, storage_options)` on a 15Gb csv file

with adlfs==0.2.4 normal speed and low memory usage
with adlfs==0.3 much slower and high memory usage

I think is related to azure-storage-blob >=v12 and I thought it would be important for you to be aware. with read_parquet didn't find issues.

Any idea? How to best report this upstream (azure)?

hayesgb:
Thanks for reporting the issue.  Can you provide some specifics on what was observed for both 0.2.4 and 0.3.0 so I can attempt to reproduce?

mgsnuno:
Yes sure:

- create a big csv file in azure storage blob
- use dask.distributed/dask_kubernetes to create a cluster
- `dd.read_csv(path, storage_options).persist()` that file

looking at the dask.distributed dashboard you should be able to see how much longer it takes to read the file. Going through the dask.distributed `Profile` you can see that an unusual amount of time is being spent on an operation part of python.core `ssl`.

If you want me to say how I created a dask_kubernetes cluster in azure I can expand on it.

andersbogsnes:
Seems like it is related to this issue: 
Azure/azure-sdk-for-python#9596

mgsnuno:
Closing for now, stopped having the issue using the same code and storage as before, either something changed internally in azure or in the library. Will reopen if reapers. Thank you.

mgsnuno:
Issue is still there, I was just being lucky with one specific file.

mgsnuno:
@hayesgb the biggest issue I experience with the latest version of adlfs while read/writing to/from csv is the high memory usage, high meaning that writing a 200Mb dataframe, with multiple partitions, can take upwards of 20Gb of RAM.
Not sure why azure-storage-blob >=12 causes the memory usage increased so much.
Any idea?

mgsnuno:
@hayesgb what can I do to help you debug this issue?
It still exists with the latest version of adlfs, if adlfs>0.3.3 then writing to an azure blob container with adlfs incurs much higher memory usage (50-100% more)

anders-kiaer:
@mgsnuno I have seen a similar problem as you (however in my case it was only a read operation, and of a `.parquet` file - the performance regression when going `>=0.3` was 10x+ in runtime).

> what can I do to help you debug this issue?

[...]

Issue #241, opened 2021-06-08
`AzureBlobFile._fetch_range` fetches too many bytes?


TomAugspurger:
Right now `_fetch_range` is implemented as https://github.com/dask/adlfs/blob/96958331854c5a67917d4c9a802a62720d091374/adlfs/spec.py#L1730-L1748. I think that's mixing up the meaning of `end` from fsspec and `length` from `azure.storage.blob`. If I fetch a with a non-zero start, I read past the `end`:

```python
In [32]: fs = adlfs.AzureBlobFileSystem("daymeteuwest")

In [33]: blob = fs.open("daymet-zarr/annual/hi.zarr/.zmetadata")

In [34]: len(blob._fetch_range(5, 10))
Out[34]: 10
```

IIUC, the intent is to read from bytes `5` to `10`, so the length should be `5`.

hayesgb:
Thanks @TomAugspurger .  You are correct.  Submitting fix now.

hayesgb:
closed with #247

PR #247, opened 2021-06-09
Added fixes for handling fetch_range extending beyond length of the file

hayesgb:
Fix for #241

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -11,13 +11,18 @@ import os
 import warnings
 import weakref
 
-from azure.core.exceptions import ResourceNotFoundError, ResourceExistsError
+from azure.core.exceptions import (
+    HttpResponseError,
+    ResourceNotFoundError,
+    ResourceExistsError,
+)
 from azure.storage.blob._shared.base_client import create_configuration
 from azure.datalake.store import AzureDLFileSystem, lib
 from azure.datalake.store.core import AzureDLFile, AzureDLPath
 from azure.storage.blob.aio import BlobServiceClient as AIOBlobServiceClient
 from azure.storage.blob import generate_blob_sas, BlobSasPermissions
 from azure.storage.blob.aio._list_blobs_helper import BlobPrefix
+from azure.storage.blob._generated.models._models_py3 import StorageError
 from azure.storage.blob._models import BlobBlock, BlobProperties, BlobType
 from fsspec import AbstractFileSystem
 from fsspec.asyn import (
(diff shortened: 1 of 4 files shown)


adlfs/spec.py
```diff
@@ -1769,20 +1772,23 @@ def connect_client(self):
                 f"Unable to fetch container_client with provided params for {e}!!"
             )
 
-    async def _async_fetch_range(self, start: int, end: int, **kwargs):
+    async def _async_fetch_range(self, start: int, length: int = None, **kwargs):
```
anders-kiaer, line 1775: This function is ultimately overriding [`_fetch_range` in `fsspec.spec.AbstractBufferedFile`](https://github.com/intake/filesystem_spec/blob/a5fa9babfe4583cf21932bd18bdae8a6e13cc3a7/fsspec/spec.py#L1445-L1447) right? If so, I guess the function signature needs to be kept `start` and `end` here, since that is what `fsspec` assumes?

adlfs/spec.py
```diff
@@ -1769,20 +1772,23 @@ def connect_client(self):
...
+        length: int
+            Length to download
         """
+        if length is not None:
+            if start + length > self.size:
+                length = self.size - start
         async with self.container_client:
             stream = await self.container_client.download_blob(
-                blob=self.blob, offset=start, length=end
+                blob=self.blob, offset=start, length=length
```
anders-kiaer, line 1791: ...and then this line instead simply becomes
```suggestion
                blob=self.blob, offset=start, length=end-start
```

anders-kiaer (commented):
Thanks for a great package :tada: I'm new to both `adlfs` and `fsspec`, but we had a use case which apparently `adlfs` solves brilliantly.

**Context:** We have been facing the following challenge: Downloading a very small subset of columns from a large `.parquet` file has very poor performance on `adlfs>=0.3`. For `adlfs<0.3` the performance is good.

At least in my case, I've pinned it down to the same thing as pointed out in #241: More bytes than necessary are downloaded from blob storage. 

I have confirmed that changing the input to Azure's `download_blob` from `length=end` to `length=end-start` gives same performance also on `adlfs>=0.3` in my case. I.e. I would guess this PR might also fix #57 at the same time.

adlfs/spec.py
hayesgb: Thanks @anders-kiaer.  I appreciate the feedback, and the confirmation that this issue resolves the challenge cited in #57.

I'll align the params in fetch_range to fsspec.AbstractBufferedFile, but I also want to account for the situation where end > self.size, and also for the eventuality that length is None, which is valid for the Azure SDK.  Can you take a look at this branch and provide feedback on performance?

anders-kiaer: Good point on the `end=None` case (and transforming that into `length=None` - that should be handled if `None` is valid input in the `fsspec` `_fetch_range` function.
> Can you take a look at this branch and provide feedback on performance?


 I'll add some quick benchmarks I did based on this branch.

anders-kiaer (commented):
I took a quick benchmark using a 2.2 GB `.parquet` file in Azure blob storage (106145 rows x 6837 columns), 
and timed how much time it took to extract one column.

Results:
* `adlfs==0.2.4` + `azure-storage-blob==2.1.0`: 0.68 ± 0.11 seconds
* `adlfs==0.7.6` + `azure-storage-blob==12.8.1`: 47.1 ± 1.22 seconds
* `adlfs==(this branch)` + `azure-storage-blob==12.8.1`: 0.64 ± 0.09 seconds

So that is a decent 99% reduction in execution time compared to `0.7.6`, and also slightly faster (however well within 1σ though) than `0.2.4` in my unofficial benchmark + with my specific `.parquet` file. The performance improvement I guess would (obviously) depend on where in the blob/file you want to extract from (extracting things from the middle of the blob/file would see a huge performance benefit after this PR, extracting from the end of the file will see the same performance as before the PR).

PR #250, opened 2021-06-28
Thec0dewriter feature default azure credential support

hayesgb:


isidentical:
Isn't this PR backwards incompatible? In the past when you had only `account_name`, it was using anonymous auth and now it is using default credentials which is a completely different behavior unless you use the new `anon` option. I believe `anon` should be defaulted to `True`.

Issue #254, opened 2021-07-15
anonymous data access behavior change


lostmygithubaccount:
Noticed the comment below in the PR. It seems like #250 introduced breaking behavior where anonymous data access cannot be done by default.

This broke our example in Azure/azureml-examples, coinciding with the pypi release of adlfs: https://github.com/Azure/azureml-examples/actions?page=2&query=workflow%3Apython-sdk-tutorial-using-dask

Was this intended? Can the previous behavior be restored?

--

Isn't this PR backwards incompatible? In the past when you had only `account_name`, it was using anonymous auth and now it is using default credentials which is a completely different behavior unless you use the new `anon` option. I believe `anon` should be defaulted to `True`.

_Originally posted by @isidentical in https://github.com/dask/adlfs/issues/250#issuecomment-880601010_

hayesgb:
Tagging @isidentical, [thec0dewriter](https://github.com/thec0dewriter), @TomAugspurger and @martindurant  for a broader discussion.

I'm open to either serving as the default behavior.  Originally, the intent was to align to the behavior of s3fs for consistency.  This explicitly implements the anon keyword, with a default value of `False` is consistent with both [s3fs](https://github.com/dask/s3fs/blob/562ed04dcc508263541a5f94fe090ef8efc1372e/s3fs/core.py#L95).

However, [gcsfs](https://github.com/dask/gcsfs/blob/07ef99ab22789bdeddea792e4322547691c0fcd4/gcsfs/core.py#L95) does not implement the `anon` kw, but instead tries a variety of approaches as well as accepting a value of anon for token.

 Assuming we want to move toward consistent behavior, what are the group's thoughts on a preferred behavior?  Should this move in the direction of s3fs or gcsfs?

TomAugspurger:
I don't have a strong preference, though the previous behavior of providing only the account name implying anonymous access seems sensible.

martindurant:
The difference between s3fs and gcsfs is one of historical convenience: whereas botocore takes case of finding various auth possibilities fo S3, leaving anon=True/False as the only realistic option, gcsfs has to manage the auth method itself, and "auto" may fail.

cc @thec0dewriter (I think the OP was supposed to tag instead of link)

isidentical:
I'd also prefer an immediate release with `anon=False` and [yanking](https://pypi.org/help/#yanked) of the previous one considering a mismatch would result with a lot of confusion.

hayesgb:
[...]

Issue #226, opened 2021-04-27
Azure ManagedIdentityCredential


rhjmoore:
I have been developing a project for a client using Dask. One key element is accessing data from their Azure Data Lake Storage. Many routes are available to authenticate to ADLS, but unfortunately the route preferred at this client (and often by Enterprise clients) is the ManagedIdentityCredential. This authentication mechanism doesn't appear to work with Dask.

https://docs.microsoft.com/en-us/python/api/azure-identity/azure.identity.managedidentitycredential?view=azure-python

Is there anything stopping it from working / being implemented through storage_options, whereby we simply present a valid credential (or perhaps a function so that each machine in a cluster can request the ManagedIdentityCredential associated with that machine)?

TomAugspurger:
Moved to adlfs. @rhjmoore please let us know if you think there's still an issue.

rhjmoore:
Apologies for the delay - holidays and now access problems to the cluster. I will respond within a week.

hayesgb:
@rhjmoore — This should work with the async credential, which can be found imported as
`import azure.identity.aio import ManagedIdentityCredential as AIOManagedIdentityCredential ` , following the same pattern used for fetching tokens for a ServicePrincipal

rhjmoore:
Here are the results of my tests. I'm not sure I understand the specifics of the error being reported, but I also have some thoughts on why ManagedIdentityCredential might be inherently problematic with the current structure.

    from azure.identity.aio import ManagedIdentityCredential as ManagedIdentityCredential_aio
    aio_mi_cred = ManagedIdentityCredential_aio(client_id=client_id)
    aio_abfs = AzureBlobFileSystem(**{'account_name': <account_name>, 'credential': aio_mi_cred})
    aio_abfs.ls('data')

This code functions as expected and lists the contents of the container "data". But now we try to use dask to read that data. For context, we have already initiated a connection to a local cluster.

    ddf = dd.read_parquet('abfs://data/input_data.parquet', engine='pyarrow', columns=[<column subset>],
                                                 storage_options={'account_name':<account_name>, 'credential': aio_mi_cred})
    df = ddf.compute()  # This subset of columns fits in memory

This now fails as the work is kicked off with:
[...]

PR #262, opened 2021-08-11
Enable DefaultAzureCredential with kw anon=False and no explicit credentials provided

hayesgb:
Tagging @isidentical @dylanfprice and @rhjmoore in this PR

This is intended to enable use of DefaultAzureCredential, with an anon parameter.

Based on #254  #226 #201 

in this PR, the default value of anon is set to TRUE, so behavior the default behavior if no credentials are explicitly provided will be to attempt to login to a public account.  If no explicit credentials are provided, and `anon=False` then it should attempt to pull credentials using `DefaultAzureCredential`

It would be great to get feedback on this, since I know there's been interest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 **Change Log**
+v2021.08.1
+------
+- Added DefaultAzureCredential as an authentication method.  Activated if anon is False, and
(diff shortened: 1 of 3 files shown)


dylanfprice:
Seems reasonable thanks for adding!

PR #275, opened 2021-09-14
Implement support for fsspec callbacks in put_file/get_file

### LINKED ISSUE #276: Implement support for fsspec callbacks
fsspec recently adopted a new feature for allowing clients to pass callbacks to track progress of methods like get_file/put_file, but adlfs doesn't support it yet. https://github.com/intake/filesystem_spec/issues/668

isidentical:
This patch adds support for [fsspec callbacks](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.callbacks.Callback). It also removes some load from the `get_file`:
- There is a completely redundant `ls()` call it is making on every `get_file` operation, which is very costful
- It also tries to do an `isdir()` check, which is also very costful (if not cached) when considering it is done for every `get_file` especially on Azure, since there are no directories per se.  

Resolves #276

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1525,7 +1525,9 @@ class AzureBlobFileSystem(AsyncFileSystem):
             raise FileNotFoundError
         return list(sorted(out))
 
-    async def _put_file(self, lpath, rpath, delimiter="/", overwrite=False, **kwargws):
+    async def _put_file(
+        self, lpath, rpath, delimiter="/", overwrite=False, callback=None, **kwargws
+    ):
         """
         Copy single file to remote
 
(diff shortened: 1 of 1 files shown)


isidentical:
CC: @TomAugspurger @hayesgb

hayesgb:
Thanks!

rubenpraets:
@isidentical In your change you link to a comment https://github.com/Azure/azure-sdk-for-python/issues/11419#issuecomment-628143480 explaining how to make a callback, but you don't follow it all the way. In particular this comment includes a check to see if current is None, which you omitted in your change. Why is that exactly? I'm asking because this is causing us trouble at the moment, please also see https://github.com/iterative/dvc/issues/7704.

Issue #284, opened 2021-10-25
`IsADirectoryError` exception when getting a directory


gabrielmbmb:
**What happened**: `IsADirectoryError` raised while trying to download a directory from Azure Blob Storage to local file system using `AzureBlobFileSystem.get` method with `recursive=True`.

**What you expected to happen**: the remote directory is downloaded without raising any error.

**Minimal Complete Verifiable Example**:

```python
import os

from adlfs import AzureBlobFileSystem

fs = AzureBlobFileSystem(
    account_name=os.getenv("ABS_ACCOUNT_NAME"),
    account_key=os.getenv("ABS_ACCESS_KEY")
)

fs.get(
    "mlops-scope/dataset/082515e9-44eb-47e0-97f5-1fdaa762953b/raw/test.parquet/", 
    "/tmp/test.parquet/", 
    recursive=True
)
```

**Anything else we need to know?**:

This is the traceback I got:

```
Traceback (most recent call last):
  File "/home/gabriel.martin/ScopeClassifier/test.py", line 13, in <module>
    fs.get(
  File "/home/gabriel.martin/ScopeClassifier/.venv/lib/python3.9/site-packages/fsspec/asyn.py", line 88, in wrapper
    return sync(self.loop, func, *args, **kwargs)
  File "/home/gabriel.martin/ScopeClassifier/.venv/lib/python3.9/site-packages/fsspec/asyn.py", line 69, in sync
    raise result[0]
  File "/home/gabriel.martin/ScopeClassifier/.venv/lib/python3.9/site-packages/fsspec/asyn.py", line 25, in _runner
    result[0] = await coro
  File "/home/gabriel.martin/ScopeClassifier/.venv/lib/python3.9/site-packages/fsspec/asyn.py", line 436, in _get
    return await _run_coros_in_chunks(
  File "/home/gabriel.martin/ScopeClassifier/.venv/lib/python3.9/site-packages/fsspec/asyn.py", line 210, in _run_coros_in_chunks
    results.append(await coro)
  File "/usr/lib/python3.9/asyncio/tasks.py", line 614, in _wait_for_one
    return f.result()  # May raise f.exception().
  File "/home/gabriel.martin/ScopeClassifier/.venv/lib/python3.9/site-packages/adlfs/spec.py", line 1629, in _get_file
    with open(lpath, "wb") as my_blob:
IsADirectoryError: [Errno 21] Is a directory: '/tmp/test.parquet'
```

**Environment**:

- Dask version:
[...]

PR #285, opened 2021-10-25
Skip directory when recursive get

gabrielmbmb:
As described in #284, `IsADirectoryError` is raised when trying to get recursively a directory from Azure Blob Storage. This PR add a check to `_get_file` method, which avoids opening a directory with `open`.


MattWenham:
Thanks. I was having the same issue, but on Windows it gives a permission denied error.

When should this propagate up to the main Pypi release? I assume it should make it into 2021-11-0 ?

gabrielmbmb:
In the meantime a new release of the package is published to PyPi, you can install the package with the latest changes directly from the Git repository.

```shell
pip install https://github.com/fsspec/adlfs.git
```

Issue #103, opened 2020-09-21
Feature request: create docs


raybellwaves:
See https://s3fs.readthedocs.io/en/latest/index.html

jhamman:
+1 to this. If there's anything I can do to help make this happen, just let me know.

PR #298, opened 2022-02-06
Start docs

TomAugspurger:
Closes https://github.com/fsspec/adlfs/issues/103

Needs to be fleshed out, but this is a start. Docs will be built in GitHub actions and hosted at http://fsspec.github.io/adlfs

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,0 +1,14 @@
+# API Reference
+
+`adlfs.AzureBlobFileSystem` provides an interface for Azure Blob Storage.
+
+```{eval-rst}
+.. autoclass:: adlfs.AzureBlobFileSystem
+    :show-inheritance:
+    :members:
+
+.. autoclass:: adlfs.AzureBlobFile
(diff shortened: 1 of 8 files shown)


docs/api.md
```diff
@@ -0,0 +1,14 @@
+# API Reference
+
+`adlfs.AzureBlobFileSystem` provides an interface for Azure Blob Storage.
```
TomAugspurger, line 3: I'm not sure about the value of this page, at least `:members:` bit which just copies in stuff from upstream (warnings and all)

jhamman: Curious what others think but I'd lean toward referencing the core filesystem interfaces like is done in gcsfs: https://github.com/fsspec/gcsfs/blob/main/docs/source/api.rst

hayesgb:
Thanks for starting this @TomAugspurger .  Closing with fixes for CI in #309

TomAugspurger:
Thanks for finishing it up!

PR #305, opened 2022-04-05
Don't drop paths containts globs

Antaxify:
The list comprehension in _expand_path drops paths that end with a '*'. As far as I understood the fsspec interface, this should be supported, so I guess this is a bug in the implementation of adlfs.

Let me know what you think. :)


hayesgb:
Thanks for digging in here.

CI tests run against Azurite, which Microsoft updated 6 weeks ago.  I'll dig into this failure mode, then I'll circle back to the question of other topics.

Antaxify:
@hayesgb did you find time to look into this?

Antaxify:
Anything I can do to get this merged?

hfurkanvural:
Any update or progress on this?

PR #307, opened 2022-04-12
Update python

hayesgb:
Updated Python versions, adds isort, and pins azure-storage-blob<12.10.0 to handle errors associated with azurite testing.


skshetry:
Hey, @hayesgb, any reason why adlfs is dropping 3.7 this early? Does adlfs only want to support the last 3 Python versions?
Also see discussion on fsspec side: https://github.com/fsspec/filesystem_spec/issues/912.

PR #309, opened 2022-04-12
Docs2

hayesgb:
Replaces #298 after fixing CI

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,35 @@
+name: Docs
+
+on:
+  push:
+    branches:    
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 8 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - run: python -m pip install -e .[docs]
+    - uses: ammaraskar/sphinx-action@v1
+      with:
+        docs-folder: "docs/"
+    - name: Commit documentation changes
+      run: |
+        git clone https://github.com/fsspec/adlfs.git --branch gh-pages --single-branch gh-pages
```
TomAugspurger, line 20: FYI @hayesgb  you might need to manually create a `gh-pages` branch before this will work.

Issue #286, opened 2021-10-28
RuntimeError : Failed to upload blockstring longer than 2147483647 bytes


manjuransari-zz:
I am running below code: 

```
from gensim.models import FastText
from gensim.test.utils import common_texts
import fsspec

model = FastText(size = 300, window=3, min_count=1)
model.build_vocab(sentences=common_texts)
model.train(sentences=common_texts, total_examples=len(common_texts), epochs=10)

fsspec_handle = fsspec.open('abfss://examplePickle.model', mode = "wb")
with fsspec_handle.open() as f:
    model.save(f)
```

This is throwing the file size error. On investigating I could find that in some scenario, the buffer size is very large. There is not check here to check the size which is throwing Failed to upload blockstring longer than 2147483647 bytes. Below method is being used to write to blob.

https://github.com/dask/adlfs/blob/bc6e7498af806fe851b4c0157612033b46c03176/adlfs/spec.py#L1929

PR #311, opened 2022-04-13
Fix overflow error while uploading data greater than 2GB

kamalsharma2:
This fixes issue #286 which throws runtime error while uploading data greater than 2GB, the given fix is to create smaller chunks of data from the input data.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1947,7 +1947,6 @@ class AzureBlobFile(AbstractBufferedFile):
 
         """
         data = self.buffer.getvalue()
-        print("Just a test")
         length = len(data)
         block_id = len(self._block_list)
         block_id = f"{block_id:07d}"
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1930,9 +1930,9 @@ class AzureBlobFile(AbstractBufferedFile):
     def get_chunks(self, data, chunk_size=1024**3):
         start = 0
         while start < len(data):
-            end = min(start + chunk_size - 1, len(data) - 1)
+            end = min(start + chunk_size , len(data))
             yield data[start:end]
-            start = end + 1
+            start = end
 
     async def _async_upload_chunk(self, final: bool = False, **kwargs):
         """
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1927,10 +1927,11 @@ class AzureBlobFile(AbstractBufferedFile):
 
     _initiate_upload = sync_wrapper(_async_initiate_upload)
 
-    def get_chunks(self, data, chunk_size=1024**3):     #Keeping the chunk size as 1 GB
+    def _get_chunks(self, data, chunk_size=1024**3):  # Keeping the chunk size as 1 GB
         start = 0
-        while start < len(data):
-            end = min(start + chunk_size , len(data))
+        length = len(data)
+        while start < length:
+            end = min(start + chunk_size, length)
             yield data[start:end]
             start = end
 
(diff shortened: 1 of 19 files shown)


kamalsharma2:
@hayesgb can you review this?

adlfs/spec.py
```diff
@@ -1927,6 +1927,13 @@ async def _async_initiate_upload(self, **kwargs):
 
     _initiate_upload = sync_wrapper(_async_initiate_upload)
 
+    def get_chunks(self, data, chunk_size=1024**3):     #Keeping the chunk size as 1 GB
```
TomAugspurger, line 1930:
```suggestion
    def _get_chunks(self, data, chunk_size=1024**3):     #Keeping the chunk size as 1 GB
```

Probably should keep this private until there's user demand for it. Will need to change where it's stored used below.

adlfs/spec.py
```diff
@@ -1927,6 +1927,13 @@ async def _async_initiate_upload(self, **kwargs):
...
+        start = 0
+        while start < len(data):
```
TomAugspurger, line 1932: Can you move this `len` and the `len` below outside of the while loop?

adlfs/spec.py
```diff
@@ -1944,13 +1951,14 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
...
         if self.mode == "wb":
             try:
-                async with self.container_client.get_blob_client(blob=self.blob) as bc:
-                    await bc.stage_block(
-                        block_id=block_id,
-                        data=data,
-                        length=length,
-                    )
-                self._block_list.append(block_id)
+                for chunk in self.get_chunks(data):
```
TomAugspurger, line 1954: Does it make sense to put the `for` inside the `async with `block below, so that we don't need to call `bc.__close__` multiple times?

adlfs/spec.py
```diff
@@ -1944,13 +1951,14 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
...
+                    async with self.container_client.get_blob_client(blob=self.blob) as bc:
+                        await bc.stage_block(
+                            block_id=block_id, data=chunk, length=len(chunk),
+                        )
+                        self._block_list.append(block_id)
+                        block_id = len(self._block_list)
+                        block_id = f"{block_id:07d}"
```
TomAugspurger, line 1961: What does this change do?

TomAugspurger (commented):
Thanks @kamalsharma2. Thoughts on adding a unit test for this? If it's reproducible with azuretie and doesn't take took long to run then I think we should add one.


I'm not sure what the CI error is since I can't view the output on Azure Pipelines. Is this something you can control @hayesgb?

adlfs/spec.py
kamalsharma2: This is to update the ```block_id``` with a new value for the next chunk.

kamalsharma2:
@TomAugspurger I've made changes as per your suggestions, the tests seem to be failing for some unrelated reason, please take a look. Thanks.

TomAugspurger:
I think we need @hayesgb or some admin on the https://dev.azure.com/fsspec/ org to flip a switch to make the logs public. I think https://docs.microsoft.com/en-us/azure/devops/organizations/public/make-project-public?view=azure-devops are the relevant docs.

TomAugspurger:
@kamalsharma2 do you have thoughts on adding a unittest for this?

kamalsharma2:
@TomAugspurger we can add a unittest but we'll be reading data > 2GB in size so it might take long to complete. What are your thoughts?

hayesgb:
@kamalsharma2 -- Apologies for the delay on getting to this.

@TomAugspurger -- I've switched the Azure Pipeline logs to public as suggested.

PR #321, opened 2022-05-19
Fix info() performance

### LINKED ISSUE #248: performances of info method
The function _info_ can be super slow when the blob storage contains many blobs.
Indeed, the function is calling _ls_ on the parent directory.

```python
out = await self._ls(
    self._parent(path), invalidate_cache=invalidate_cache, **kwargs
)
```
When working on large blob storage _ls_ can takes ages.

In my scenario, I'm just calling the dask function "read_parquet" which calls  _isdir_ which calls the _info_ methods.

Is It a known issue?
For now, my fix is to create an "extra directory" that makes _ls_ faster.

Thanks,

**Environment**:

- Dask version: 2021.4.1
- Python version: 3.8
- Operating System: win 10. Linux

ofrasergreen:
This reorders the behavior of fs.info() so that it now searches the
parent directory for the entry after trying to find the item directly.
This is because scanning the parent directory is O(N) with the number of
items in the directory.

Fixes #248

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -645,12 +645,6 @@ class AzureBlobFileSystem(AsyncFileSystem):
         else:
             invalidate_cache = False
         path = self._strip_protocol(path)
-        out = await self._ls(
-            self._parent(path), invalidate_cache=invalidate_cache, **kwargs
-        )
-        out = [o for o in out if o["name"].rstrip("/") == path]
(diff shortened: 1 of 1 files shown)


astrowonk:
Thanks @ofrasergreen - this would fix #349 because it appears `_info` is also called during `.open`.

PR #322, opened 2022-06-06
Pass kwargs through to `_glob`

dhirschfeld:


dhirschfeld:
I was surprised when passing `detail=True` didn't do anything.

PR #323, opened 2022-06-16
Update `AzureBlobFileSystem` docstrings

alvarobartt:
I was checking `AzureBlobFileSystem` docstrings in order to play around with it and I found out that in the docstrings it says that one authentication method allows you to provide not just the `account_name` and `account_key` but also the `container_name`.

So on, I removed the `container_name` arg from the docstring as when calling `abfs.ls("")` the output is not the listing of the container's directory, but the listing of all the containers in the account, so the `container_name` arg won't have any use there, as in order to list all the blobs of a container we could just `abfs.ls("container_name")`.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -359,7 +359,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
     Authentication with an account_key
 
-    >>> abfs = AzureBlobFileSystem(account_name="XXXX", account_key="XXXX", container_name="XXXX")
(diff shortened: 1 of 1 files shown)


alvarobartt:
@TomAugspurger as #324 has already been merged, could we also merge this small PR? Note that the unit tests will succeed on merge as those are already solved in `master` and this PR is just changing a docstring. Thanks!

TomAugspurger:
Thanks!

PR #324, opened 2022-06-20
Fix unit tests as an `AssertionError` was triggered while listing the metadata

alvarobartt:
When I created a PR for `adlfs` at https://github.com/fsspec/adlfs/pull/323 I found out that the unit tests triggered by `azure-pipelines.yml` were failing not just for my PR but for all the other PRs, and since mine just changed a docstring, that means that the unit tests were failing before.

So I did a little bit of research and found out that in the stable version of Dask as explained at https://docs.dask.org/en/stable/generated/dask.dataframe.to_parquet.html#dask.dataframe.to_parquet, you need to manually set the value for `write_metadata_file` when calling `.to_parquet` so that not just the parquet files are written, but also the metadata (both `_metadata` and `_common_metadata` directories).

So on, adding the parameter `write_metadata_file=True` in all the `.to_parquet` calls fixes the issue, and the tests are passing again.

### DIFF
--- a/adlfs/tests/test_spec.py
+++ b/adlfs/tests/test_spec.py
@@ -939,6 +939,7 @@ def test_dask_parquet(storage):
             "{}://test/test_group.parquet".format(protocol),
             storage_options=STORAGE_OPTIONS,
             engine="pyarrow",
+            write_metadata_file=True,
         )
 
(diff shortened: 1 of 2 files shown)


alvarobartt:
Hi @hayesgb, could you please check this PR? Thanks :hugs:

TomAugspurger:
Thanks!

PR #326, opened 2022-06-24
Parse AzureBlobFileSystem account_name in URI

GuillaumeDesforges:
Before this PR, one needs to manually specify `account_name` to `storage_options` when opening a file from Blob Storage with fsspec.

```python
import fsspec

# ok
with fsspec.open("abfs://test@some_account_name.dfs.core.windows.net/some_file", account_name="") as f:
   ...

# fails
with fsspec.open("abfs://test@some_account_name.dfs.core.windows.net/some_file") as f:
   ...
```

However, the account name can be part of the URI, as specified by [Azure's doc](https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri#uri-syntax).

This PR allows the user not to specify `account_name` to `storage_options` when it is in the URI

```python
import fsspec

# ok
with fsspec.open("abfs://test@some_account_name.dfs.core.windows.net/some_file", account_name="") as f:
   ...

# ok
with fsspec.open("abfs://test@some_account_name.dfs.core.windows.net/some_file") as f:
   ...
```

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -7,6 +7,7 @@ import asyncio
 import io
 import logging
 import os
+import re
 import warnings
 import weakref
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


hayesgb:
Thanks for the contribution!

Issue #270, opened 2021-08-26
DefaultAzureCredential authentication path throws NotImplementedError on Windows 10


edwardlonghurst:
**What happened**:
On Windows 10:
Including ```'anon':False``` in the storage_options throws ```NotImplementedError``` on ```asyncio.get_child_watcher().attach_loop(self.loop)```

**What you expected to happen**:
DefaultAzureCredentials are used to authenticate

**Minimal Complete Verifiable Example**:
On Windows 10:
```python
import pandas as pd
url = "abfs://..."
pd.read_parquet(url, storage_options={"anon":False, "account_name":"..."})
```

**Anything else we need to know?**:
Providing the credentials directly (thus avoiding the call to get_child_watcher) works fine:

```python
from azure.identity.aio import DefaultAzureCredential as AIODefaultAzureCredential
import pandas as pd
url = "abfs://..."
pd.read_parquet(url, storage_options={"anon":False, "account_name":"...", 'credential':AIODefaultAzureCredential()})
```

**Environment**:

- Dask version: 2021.8,1
- adlfs version: 2021.8.2
- fsspec version: 2021.7.0
- Python version: 3.8.8
- Operating System: Windows 10
- Install method (conda, pip, source): pip

gabrieldaiha:
Man, I've faced the same problem here.

My workaround was (you can see it in my fork, i'm preparing it for PR):

In get_default_azure_credential, I removed the asyncio part (Windows does not have get_child_watcher implemented, this function is Unix only):

 https://github.com/gabrieldaiha/adlfs/blob/9d5283f5630e500f71d1ab91e7cc24e8c6592afb/adlfs/spec.py#L532-L539

I moved the get_default_azure_credential call from do_connect function to the __init__ function, like get_service_principal_credential does.
I added a new weakref.finalize to async credentials too, in order to avoid Unclosed Client Session ERROR in the exit of the program

https://github.com/gabrieldaiha/adlfs/blob/9d5283f5630e500f71d1ab91e7cc24e8c6592afb/adlfs/spec.py#L455-L466

PeterFogh:
I think this issue is fixed by #328, thus it can be closed.

PR #328, opened 2022-06-30
Fixing Default Azure Credential bug in Windows throwing NotImplementedError

gabrieldaiha:
This pull request aims to solve the #270 issue about throwing NotImplementedError when trying to use anon=False in Windows.
This error occurs because asyncio.get_child_watcher is Unix only. 

The workaround was:

In get_default_azure_credential, I removed the asyncio part (Windows does not have get_child_watcher implemented, this function is Unix only):

https://github.com/gabrieldaiha/adlfs/blob/223d6683e35757b25671b42dd7cfdc67df08df18/adlfs/spec.py#L531-L548

I moved the get_default_azure_credential call from do_connect function to the init function, like get_service_principal_credential does. Finally, I added a new weakref.finalize to async credentials too, in order to avoid Unclosed Client Session ERROR in the exit of the program:

https://github.com/gabrieldaiha/adlfs/blob/223d6683e35757b25671b42dd7cfdc67df08df18/adlfs/spec.py#L454-L465


hayesgb:
I suspect the linter is complaining b/c of the single linespace after the new function in `utils.py`, but you should be able to run `pre-commit run --all-files` to fix the lint issues.

hayesgb (approved):

PR #329, opened 2022-07-23
Limit concurrency

hayesgb:
Picking up of #288

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -7,6 +7,7 @@ import asyncio
 import io
 import logging
 import os
+import re
 import warnings
 import weakref
 from datetime import datetime, timedelta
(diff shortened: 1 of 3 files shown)


TomAugspurger:
Thanks for picking this up!

One thing I've learned about since starting this PR, azure-storage-blob already uses the `max_concurrency` keyword for something that's *slightly*  different. The `BlobClient.upload_blob` and `BlobClient.download_blob` use `max_concurrency` to control the number of threads / concurrent tasks to run. It's similar, but slightly difference since it controls concurrency / parallelism on a single blob operation. This PR was targeting operations on many blobs.

We might want to consider a different keyword (max_clients? max_blob_clients?) so that we can pass through `max_concurrency` for parallelism on single-blob operations.

hayesgb:
Hey @TomAugspurger -- That sounds like a great idea.

Need to figure out why these are failing here -- they were passing in my local environment.

PR #330, opened 2022-07-23
Fixed weakref async credential close when self.credential is None

gabrieldaiha:
The PR #328 included a weakref to close async credential, to avoid the Unclosed Client Session warning.

This PR aims to fix the case that we can have self.credential = None after passing the self.do_connect.

Lines 470-471 of spec.py is the fix:
https://github.com/gabrieldaiha/adlfs/blob/d2a5be68af7934025f884d2dc523f1708738c6d7/adlfs/spec.py#L470-L471


gabrieldaiha:
Tests are claiming an error about resource name size that it does not have relationship with this PR.
@hayesgb anything happened with yesterday's MR in these tests? 
- test_fetch_entire_blob
- test_fetch_first_half
- test_fetch_middle

hayesgb:
Not that I'm aware of, but I'll check.  Its interesting that its only failing in DEV

gabrieldaiha:
> Not that I'm aware of, but I'll check. Its interesting that its only failing in DEV

Any news?

kyleknap:
This looks now handled based on this code path: https://github.com/fsspec/adlfs/blob/74ed2be8fc02cd2216ca1e6cc04d2fed98ddc017/adlfs/spec.py#L381-L382 Closing PR.

PR #331, opened 2022-08-08
abfs: support blob storage versioning

pmrowla:
Implements `AzureBlobFileSystem` support for [blob versioning](https://docs.microsoft.com/en-us/azure/storage/blobs/versioning-overview) equivalent to what is currently supported in s3fs

- Filesystem must be configured with `version_aware=True` (and container must have versioning enabled)
- `version_id` property returned in fs read methods that return file details (i.e. `find`, `ls`, `info`)
    - Defaults to `None` when filesystem or container is not versioning-aware
- For file read methods specific blob version ID can be specified via `version_id` kwarg or via the requested path using `versionid` query parameter
    - path example: `abfs://my_container/path/to/versioned_file?versionid=some_version_id`
    - `versionid` used in path instead of `version_id` so paths are consistent with the Azure REST API [URI format](https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob#uri-parameters). This is consistent with the s3fs implementation which uses `version_id` in the fsspec python API and `VersionId` in paths (to match the AWS URI format)

Regarding testing:
- Version-aware filesystem methods have been manually tested against actual Azure containers with versioning enabled
- Azurite does not support blob versioning (it will ignore any versioning related API arguments)
- Version-aware filesystem methods are currently tested with mocked Azure SDK calls to ensure that we at least make calls with the correct arguments, and can parse/filter any returned version-aware blob properties

~~TODO:~~
- [x] figure out easiest test implementation (azurite does not currently implement any of the blob versioning functionality: https://github.com/Azure/Azurite/issues/665)

### DIFF
--- a/adlfs/tests/conftest.py
+++ b/adlfs/tests/conftest.py
@@ -59,9 +59,13 @@ def storage(host):
 
 @pytest.fixture(scope="session")
 def event_loop():
-    loop = asyncio.get_event_loop()
-    yield loop
-    loop.close()
+    policy = asyncio.get_event_loop_policy()
+    loop = policy.new_event_loop()
+    try:
+        yield loop
+    finally:
+        loop.close()
+        policy.set_event_loop(loop)
 
 
 @pytest.fixture(scope="session", autouse=True)
--- a/adlfs/utils.py
+++ b/adlfs/utils.py
@@ -1,4 +1,15 @@
-async def filter_blobs(blobs, target_path, delimiter="/"):
+from typing import Optional
+
+
+def match_blob_version(blob, version_id: Optional[str]):
+    return (
+        version_id is None and ("version_id" not in blob or blob["is_current_version"])
+    ) or (version_id is not None and blob["version_id"] == version_id)
+
+
+async def filter_blobs(
+    blobs, target_path, delimiter="/", version_id: Optional[str] = None
+):
     """
     Filters out blobs that do not come from target_path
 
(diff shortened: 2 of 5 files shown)


adlfs/tests/conftest.py
```diff
@@ -59,7 +59,7 @@ def storage(host):
 
 @pytest.fixture(scope="session")
 def event_loop():
-    loop = asyncio.get_event_loop()
```
pmrowla, line 62: `get_event_loop()` is deprecated in 3.10, and separate sessions should use their own loops

adlfs/tests/conftest.py
```diff
@@ -59,9 +59,13 @@ def storage(host):
...
-    yield loop
-    loop.close()
+    policy = asyncio.get_event_loop_policy()
+    loop = policy.new_event_loop()
+    try:
+        yield loop
+    finally:
+        loop.close()
+        policy.set_event_loop(loop)
```
pmrowla, line 68: with pytest-asyncio`mode=auto` we need to explicitly update the policy eventloop after closing otherwise `asyncio` may overwrite it: https://github.com/pytest-dev/pytest-asyncio/issues/257

efiop (approved):
LGTM, but I have a conflict of interest here 😄 (we need it for active use in dvc) Waiting for another review.

setup.py
```diff
@@ -33,7 +33,7 @@
         "azure-core>=1.7.0",
         "azure-datalake-store>=0.0.46,<0.1",
         "azure-identity",
-        "azure-storage-blob>=12.5.0",
+        "azure-storage-blob>=12.12.0",
```
pmrowla, line 36: Blob versioning related adlfs changes require using the current azure service version and pagination changes from 12.12.0:

https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/CHANGELOG.md#12120b1-2022-04-14

efiop:
@TomAugspurger @hayesgb Hey guys. 🙂  Could you please take a look at this PR when you have time? 🙏

efiop:
It feels a bit intrusive from my side, but it looks like there are no comments (I'm sure people are just busy with other stuff), old and new tests pass and we are ready to help maintain this, so I hope no one will mind me taking the liberty to move forward with this 🙂

efiop:
@TomAugspurger Thank you! 🙏

adlfs/utils.py
```diff
@@ -1,4 +1,15 @@
-async def filter_blobs(blobs, target_path, delimiter="/"):
+from typing import Optional
+
+
+def match_blob_version(blob, version_id: Optional[str]):
+    return (
+        version_id is None and ("version_id" not in blob or blob["is_current_version"])
```
efiop, line 6: version_id could actually be present but set to `None`, which will make us ignore blobs when `fs.version_aware==True`. So maybe

``suggestion
        version_id is None and (blob.get("version_id") is None or blob["is_current_version"])
```

?

efiop: Also, there is `blob["version_id"]` a line below, maybe we shouldn't assume that it is always present in `blob` too and do a `get()` instead?

pmrowla: should be fixed in bbd9483

hayesgb:
Thanks.  This looks great!

PR #332, opened 2022-08-17
Fixed strip_protocol not stripping left / when passing a full uri

nickolasrm:
When calling glob passing a path like this:

```
abfs://account.dfs.core.windows.net/container/path
# or
account.dfs.core.windows.net/container/path
```

adlfs returns an empty list. However, by debugging the code, I noticed the method
`_strip_domain` (used by glob) was returning `/container/path` instead of `container/path`,
which made adlfs to not find the files.

This pr just lstrips the output of the method and adds a test for it


hayesgb:
Thanks!

PR #337, opened 2022-08-28
gha: fix docs generation

efiop:
ammaraskar/sphinx-action requires docs/requirements.txt that was missing.

GHA now runs on PRs too, just to make sure we don't mess up docs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,6 +1,7 @@
 name: Docs
 
 on:
+  pull_request:
   push:
     branches:    
       - main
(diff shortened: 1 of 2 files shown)


efiop:
Ok, turns out docs generation was completely not operational before 🙁

efiop:
I hope you folks don't mind me messing with this directly, just want to help improve this project for everyone. 🙏

efiop:
https://fsspec.github.io/adlfs/ works now, but looks like our aim is RTD, similar to gcsfs

PR #338, opened 2022-08-29
AzureBlobFileSystem: remove socket_timeout

### LINKED ISSUE #303: AzureBlobFileSystem: socket_timeout ignored
It seems that the `socket_timeout` argument for `AzureBlobFileSystem` is silently ignored:

https://github.com/fsspec/adlfs/blob/fbe43c601e7b162b7a2bf93f7f8b699c747b70e6/adlfs/spec.py#L404

is this intended?

dtrifiro:
`socket_timeout` option is not being used

closes #303

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -63,6 +63,8 @@ VERSIONED_BLOB_PROPERTIES = [
 ]
 _ROOT_PATH = "/"
 
+_SOCKET_TIMEOUT_DEFAULT = object()
+
 
 class AzureDatalakeFileSystem(AbstractFileSystem):
     """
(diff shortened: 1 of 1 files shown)


TomAugspurger:
Thanks! Could you do this as a deprecation warning to avoid breaking code that was passing that?

Since we use `None` for the default, you'd need something like

```python
# top-level
_socket_timeout_default = object()

def __init__(self, ... socket_timeout=_socket_timeout_default, ...):
    # in __init__
    if socket_timeout is not _socket_timeout_default:
        warnings.warn("message", FutureWarning)
```

adlfs/spec.py
```diff
@@ -460,7 +458,10 @@ def __init__(
         self.location_mode = location_mode
         self.credential = credential
         self.request_session = request_session
-        self.socket_timeout = socket_timeout
+        if socket_timeout is not _SOCKET_TIMEOUT_DEFAULT:
+            warnings.warn(
+                "socket_timeout is deprecated and has no effect.", DeprecationWarning
```
TomAugspurger, line 463: A FutureWarning will be visible in more places than the DeprecationWarning.

```suggestion
                "socket_timeout is deprecated and has no effect.", FutureWarning
```

TomAugspurger (approved):

Issue #340, opened 2022-09-01
Support user account_host and check if blob size exists


KeerthiYandaOS:
* [Adlfs spec](https://github.com/fsspec/adlfs/blob/main/adlfs/spec.py#L589) has the hostname with "blob.core.windows.net" but all the URLs that we access might not have the same extension in the host, it will be helpful if we can provide an ability to accept the URL host from the user. 
* We are checking the blob size [here](https://github.com/fsspec/adlfs/blob/main/adlfs/spec.py#L895) but we are seeing the error as the "size" parameter is not found. In order to overcome this, we can check if the size attribute exists and then check the size value.

KeerthiYandaOS:
Created pull request for this issue: https://github.com/fsspec/adlfs/pull/341

PR #341, opened 2022-09-01
Support account_host and check blob size attribute exists

KeerthiYandaOS:
Addresses the issue https://github.com/fsspec/adlfs/issues/340

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -584,10 +584,16 @@ class AzureBlobFileSystem(AsyncFileSystem):
                 self.service_client = AIOBlobServiceClient.from_connection_string(
                     conn_str=self.connection_string
                 )
-            elif self.account_name:
-                self.account_url: str = (
-                    f"https://{self.account_name}.blob.core.windows.net"
-                )
+            elif self.account_name is not None:
+                if hasattr(self, 'account_host'):
+                    self.account_url: str = (
+                        f"https://{self.account_host}"
+                    )
+                else:
+                    self.account_url: str = (
+                        f"https://{self.account_name}.blob.core.windows.net"
+                    )
+                
                 creds = [self.credential, self.account_key]
                 if any(creds):
                     self.service_client = [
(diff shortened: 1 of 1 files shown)


KeerthiYandaOS:
@hayesgb Can you please review the pull request? I'm unable to access the build logs, can you provide me access or steps to get the access? Thanks.

adlfs/spec.py
```diff
@@ -584,10 +584,16 @@ def do_connect(self):
                 self.service_client = AIOBlobServiceClient.from_connection_string(
                     conn_str=self.connection_string
                 )
-            elif self.account_name:
-                self.account_url: str = (
-                    f"https://{self.account_name}.blob.core.windows.net"
-                )
+            elif self.account_name is not None:
+                if hasattr(self, 'account_host'):
```
TomAugspurger, line 588: Can you help me understand where this `account_host` attribute coming from? Is this from a subclass?

I'd rather have the `__init__` method accept something like `account_host` and make it a property on `AzureBlobFileSystem`. Then we can avoid all the hasattrs.

The interaction with the current `account_name` will be a bit complicated. Can you share an example of how account_host is being specified? Maybe we just don't allow users to pass both.

adlfs/spec.py
```diff
@@ -892,7 +898,7 @@ async def _ls(
                                                 outblobs.append(blob_)
                                             elif (
                                                 blob_["name"].count("/") == depth
-                                                and blob_["size"] == 0
+                                                and (hasattr(blob_, "size") and blob_["size"] == 0)
```
TomAugspurger, line 901: Can you share a bit about why the `blob_` would have a size attribute? Does the lack of `size` attribute indicate a "directory" marker?

adlfs/spec.py
KeerthiYandaOS: Adlfs is always appending the public blob endpoint. so having the ability to pass the account_host will help in accessing the non-public regions like gov regions. As account_host is not mandatory I was thinking of consumers to pass as required. We can also add it in __init__ and initialize it as None making as not a required property.

adlfs/spec.py
KeerthiYandaOS: Yes Tom, have seen the blob not having size if the blob type is not a file.

Issue #343, opened 2022-09-06
aiohttp not in requirements


benrutter:
Looks like the latest version removes aiohttp from requirements, but not imports, meaning that an error will be raised importing adlfs if aiohttp isn't additionally installed.

Reproduction steps:

1. Create a fresh python environment
2. Install the latest version of adlfs
3. Import adlfs

This should throw a module not found error for aiohttp.

I think the resolution for this is just to add aiohttp into the requirements? I've put in a pull request if that is the case here: https://github.com/fsspec/adlfs/pull/344

TomAugspurger:
This is fixed and released to PyPI as 2022.9.1: https://pypi.org/project/adlfs/2022.9.1/

PR #344, opened 2022-09-06
Update setup.py to include aiohttp

benrutter:
aiohttp added to requirements, related to issue 343 (https://github.com/fsspec/adlfs/issues/343)

This is a simple fix, and might not be the right approach but currently:

- The latest version of adlfs fails if aiohttp not installed, and doesn't install adlfs thorough pip (looking at the git blame, I think it may be a problem on the end of azure-storage, so could just be a problem exposed from there)

This pull request:

- Adds on aiohttp to the requirements

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
         "azure-identity",
         "azure-storage-blob>=12.12.0",
         "fsspec>=2021.10.1",
+        "aiohttp>=3.3.0",
     ],
     extras_require={
         "docs": ["sphinx", "myst-parser", "furo", "numpydoc"],
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(
         "azure-identity",
         "azure-storage-blob>=12.12.0",
         "fsspec>=2021.10.1",
-        "aiohttp>=3.3.0",
+        "aiohttp>=3.7.0",
     ],
     extras_require={
         "docs": ["sphinx", "myst-parser", "furo", "numpydoc"],


TomAugspurger:
Thanks for catching this. I didn't realize that aiohttp needed to be installed separately from azure-storage-blob, but it does (https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/CLIENT_LIBRARY_DEVELOPER.md#transport)

TomAugspurger:
@hayesgb this requires a new release I think. I'll get started on that now.

setup.py
```diff
@@ -34,6 +34,7 @@
         "azure-identity",
         "azure-storage-blob>=12.12.0",
         "fsspec>=2021.10.1",
+        "aiohttp>=3.7.0",
```
efiop, line 37: We probably should add a note here or something. Otherwise https://github.com/fsspec/adlfs/pull/335 will happen again.

efiop: cc @skshetry

dhirschfeld:
> Thanks for catching this. I didn't realize that aiohttp needed to be installed separately from azure-storage-blob, but it does ([Azure/azure-sdk-for-python@`main`/sdk/core/azure-core/CLIENT_LIBRARY_DEVELOPER.md#transport](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/CLIENT_LIBRARY_DEVELOPER.md?rgh-link-date=2022-09-06T11%3A12%3A07Z#transport))

Since I'm here wondering why `aiohttp` is required I'll add some more context:
https://github.com/Azure/azure-sdk-for-python/issues/19201

Probably does deserve a note in the `pyproject.toml`...

PR #350, opened 2022-09-27
Respect onerror='return' in AzureBlobFileSystem.cat

ErlendHaa:
Hi!

I noticed that the `AzureBlobFileSystem.cat` does not respect the onerror='return' (and 'omit'). Essentially, none-existing paths are always removed from the out set. This also causes problems upstream, e.g. in `fspec.FSmap.getitems` [1] because this method expects to filter the path's itself. The result is a `KeyError` every time the key-set includes none-existing blob.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1618,7 +1618,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
             included in the output, but the value will be bytes or an exception
             instance.
         """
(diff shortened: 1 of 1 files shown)


hayesgb:
Thanks for this contribution @ErlendHaa!  It was failing the linter.  I patch this and merged your changes in.

ErlendHaa:
Hi again! as far as I can see #353 did not include this change?

PR #352, opened 2022-10-01
Revert "Support account_host and check blob size attribute exists"

hayesgb:
Reverts fsspec/adlfs#341


KeerthiYandaOS:
@hayesgb can you please not revert the changes? We need support for non-public(blob.core.windows.net) hosts.

PR #364, opened 2022-10-27
Advanced Settings Parameters

barmettl:
This is a draft requesting feedback on how to implement more fine granular parameterization of the underlying BlobServiceClients and corresponding requests.

## Concrete motivation

We cannot accept slow feedback on failed put requests. If connection is not available the request should fail within a few seconds. This can be achieved by passing corresponding keyword arguments to the BlobService clients and the upload request.

## Possible implementations

It is probably best to implement advanced settings in general with untyped keyword arguments/dictionaries which are passed directly to underlying calls. These should correspond exactly to the keyword arguments of the azure.storage.blob classes and the user is forwarded to the documentation of the azure sdk on the usage of these parameters. 

The question is how the signature of `AzureBlobFileSystem` should look like. Possible implementation by example: 

### Subdictionary for each subclass / call

```yaml
anon: True
# all other typed arguments
kwargs:
  exclude_cli_credential: False
  exclude_environment_credential: True
# .... any other arguments 
client_kwargs:
  initial_backoff: 2
  increment_base: 1
  random_jitter_range: 1
# .... any other arguments passed to BlobServiceClient
upload_kwargs:
  timeout: 2
# .... any other arguments passed to upload.

```

I implemented this version first because it is safest with respect to unintended effects and naming conflicts. This can be extended to more requests, i.e. `download_kwargs` and others. One could also skip the `_kwargs` suffix.

### kwargs forwarding

```yaml
anon: True
# all other typed arguments
kwargs:
  exclude_cli_credential: False
  exclude_environment_credential: True
  initial_backoff: 2
  increment_base: 1
  random_jitter_range: 1
  timeout: 2
# .... any other arguments passed to BlobServiceClient and requests.

```
This is more in line with the current implementation and should also work, haven't checked though. There might be some naming conflicts.

## Design decision

Both of these approaches, and anything inbeetween would also work. This is mainly a question of taste and consistency rather than correctness, so I want to reach to @TomAugspurger and other experts to make a decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 **Change Log**
-2022.11.0
+2022.11.2
 ---------
 - Reorder fs.info() to search the parent directory only after searching for the specified item directly
 - Removed pin on upper bound for azure-storage-blob
 - Moved AzureDatalakeFileSystem to a separate module, in acknowledgement of [Microsoft End of Life notice](https://learn.microsoft.com/en-us/answers/questions/281107/azure-data-lake-storage-gen1-retirement-announceme.html})
+- Added DeprecationWarning to AzureDatalakeFilesystem
 
 
 2022.10.1
--- a/adlfs/gen1.py
+++ b/adlfs/gen1.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import, division, print_function
 
 import logging
+from warnings import warn
 
 from azure.datalake.store import AzureDLFileSystem, lib
 from azure.datalake.store.core import AzureDLFile, AzureDLPath
@@ -61,6 +62,12 @@ class AzureDatalakeFileSystem(AbstractFileSystem):
     protocol = "adl"
 
     def __init__(self, tenant_id, client_id, client_secret, store_name):
+        warn(
+            f"Microsoft has announced end of life for Azure Datalake Gen1.  {self.__class__.__name__}"
+            " will be moved to an optional install in a future release of adlfs",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         super().__init__()
         self.tenant_id = tenant_id
         self.client_id = client_id
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ setup(
     long_description=open("README.md").read() if exists("README.md") else "",
     install_requires=[
         "azure-core>=1.23.1,<2.0.0",
+        "azure-datalake-store>=0.0.46,<0.1",
         "azure-identity",
         "azure-storage-blob>=12.12.0",
         "fsspec>=2021.10.1",
@@ -37,7 +38,6 @@ setup(
     ],
     extras_require={
         "docs": ["sphinx", "myst-parser", "furo", "numpydoc"],
-        "AzureDatalakeFileSystem": ["azure-datalake-store>=0.0.46,<0.1"],
     },
     tests_require=["pytest", "docker", "pytest-mock", "arrow", "dask[dataframe]"],
     zip_safe=False,


barmettl:
@hayesgb, can you give feedback on this?

hayesgb:
Hi @barmetti.  Apologies for the delay responding here.

Normally, I would lean toward passing all params that are passed to `BlobServiceClient` method being embedded in `client_kwargs`, but given the large number of potential params, and the risk of collision between params across the different methods, I like you proposal of splitting these into `upload_kwargs`, `download_kwargs`.

My only concern (which is small, and I wouldn't block on it I think) is the possible complexity if this is used in a `dask io` operation.

```
dd.read_parquet(path,
    storage_options={
        "connection_string": CONN_STR, 
        "client_kwargs": {},
        "upload_kwargs": {},
        "download_kwargs": {},
)
```

barmettl:
OK, so we agree on nesting args for passing advanced settings parameters to the azure sdk. We use the `_kwargs` suffix to semantically distinguish these params from the rest.

Not quite sure if I correctly understood the `dask io` problem you are referring. One does not need to require keys for the additional kwargs, use empty dictionaries if they are not present. Your example:

```
storage_options={
        "connection_string": CONN_STR, 
        "client_kwargs": {},
        "upload_kwargs": {},
        "download_kwargs": {}
}
```

Would be equivalent to:

```
storage_options={
        "connection_string": CONN_STR
}
```

I will come up with an implementation of this feature. The goal is to provide full control over the azure.storage.blob SDK.

barmettl:
I checked all the azure.store.blob invocations in spec.py. There are far more than just upload/download. Continuing  upload_kwargs, download_kwargs for all methods will result in lots of duplications of the same parameter with the same meaning. For example the `timeout` argument is used in all methods. It is possible to use a different value for each method, but this almost never makes sense.

Instead, I suggest we just use the existing kwargs field to pass through any argument to azure.store.blob methods. 

This may result in conflicts. In order to avoid this I compiled a list of all arguments (excluding explicitly set arguments such as `name`) of azure.store.blob functions used in spec.py. I found, that if the same argument name is used in different methods, the description string is identical (up to formatting) and identical argument name means same type. 

Here is the full list

```yaml
api_version:  
  description: "The Storage API version to use for requests. Default value is the
    most recent service version that is compatible with the current SDK. Setting to
    an older version may result in reduced feature compatibility.  .. versionadded::
    12.2.0"
  methods: BlobServiceClient
  type: str
blob_type:
  description: The type of the blob. This can be either BlockBlob, PageBlob or AppendBlob.
    The default value is BlockBlob.
  methods: BlobClient.upload_blob
  type: azure.storage.blob.BlobType
container_encryption_scope:
  description: "Specifies the default encryption scope to set on the container and
    use for all future writes.  .. versionadded:: 12.2.0"
  methods: BlobServiceClient.create_container
  type: Any
content_settings:
  description: ContentSettings object used to set blob properties. Used to set content
    type, encoding, language, disposition, md5, and cache control.
  methods: BlobClient.upload_blob
  type: azure.storage.blob.ContentSettings
cpk:
  description: Encrypts the data on the service-side with the given key. Use of customer-provided
    keys must be done over HTTPS. As the encryption key itself is provided in the
    request, a secure connection must be established to transfer the key.
  methods: BlobClient.get_blob_properties | BlobClient.download_blob | BlobClient.upload_blob
  type: azure.storage.blob.CustomerProvidedEncryptionKey
credential:
  description: The credentials with which to authenticate. This is optional if the
    account URL already has a SAS token. The value can be a SAS token string, an instance
    of a AzureSasCredential from azure.core.credentials, an account shared access
    key, or an instance of a TokenCredentials class from azure.identity. If the resource
    URI already contains a SAS token, this will be ignored in favor of an explicit
    credential - except in the case of AzureSasCredential, where the conflicting SAS
    tokens will raise a ValueError.
  methods: BlobServiceClient
  type: Any
delete_snapshots:
  description: "Required if the blob has associated snapshots. Values include: - "only":
    Deletes only the blobs snapshots. - "include": Deletes the blob along with all
    snapshots."
  methods: ContainerClient.delete_blob
  type: str
delimiter:
  description: When the request includes this parameter, the operation returns a BlobPrefix
    element in the response body that acts as a placeholder for all blobs whose names
    begin with the same substring up to the appearance of the delimiter character.
    The delimiter may be a single character or a string.
  methods: ContainerClient.walk_blobs
  type: str
destination_lease:
  description: The lease ID specified for this header must match the lease ID of the
    destination blob. If the request does not include the lease ID or it is not valid,
    the operation fails with status code 412 (Precondition Failed).
  methods: BlobClient.start_copy_from_url
  type: Any
encoding:
  description: Encoding to decode the downloaded bytes. Default is None, i.e. no decoding.
  methods: BlobClient.download_blob | BlobClient.upload_blob
  type: str
encryption_scope:
  description: "A predefined encryption scope used to encrypt the data on the service.
    An encryption scope can be created using the Management API and referenced here
    by name. If a default encryption scope has been defined at the container, this
    value will override it if the container-level scope is configured to allow overrides.
    Otherwise an error will be raised.  .. versionadded:: 12.2.0"
  methods: BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: str
etag:
  description: An ETag value, or the wildcard character (*). Used to check if the
    resource has changed, and act according to the condition specified by the `match_condition`
    parameter.
  methods: ContainerClient.delete_blob | BlobServiceClient.delete_container | BlobClient.get_blob_properties
    | BlobClient.download_blob | BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: str
if_modified_since:
  description: A DateTime value. Azure expects the date value passed in to be UTC.
    If timezone is included, any non-UTC datetimes will be converted to UTC. If a
    date is passed in without timezone info, it is assumed to be UTC. Specify this
    header to perform the operation only if the resource has been modified since the
    specified time.
  methods: ContainerClient.delete_blob | BlobServiceClient.delete_container | BlobClient.get_blob_properties
    | BlobClient.download_blob | BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: datetime.datetime
if_tags_match_condition:
  description: "Specify a SQL where clause on blob tags to operate only on blob with
    a matching value. eg. ``"\\"tagname\\"=""my tag"""``  .. versionadded:: 12.4.0"
  methods: ContainerClient.delete_blob | BlobClient.get_blob_properties | BlobClient.download_blob
    | BlobClient.upload_blob
  type: str
if_unmodified_since:
  description: A DateTime value. Azure expects the date value passed in to be UTC.
    If timezone is included, any non-UTC datetimes will be converted to UTC. If a
    date is passed in without timezone info, it is assumed to be UTC. Specify this
    header to perform the operation only if the resource has not been modified since
    the specified date/time.
  methods: ContainerClient.delete_blob | BlobServiceClient.delete_container | BlobClient.get_blob_properties
    | BlobClient.download_blob | BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: datetime.datetime
immutability_policy:
  description: "Specifies the immutability policy of a blob, blob snapshot or blob
    version. Currently this parameter of upload_blob() API is for BlockBlob only.  ..
    versionadded:: 12.10.0     This was introduced in API version ""2020-10-02""."
  methods: BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: azure.storage.blob.ImmutabilityPolicy
include_deleted:
  description: "Specifies that deleted containers to be returned in the response.
    This is for container restore enabled account. The default value is `False`. ..
    versionadded:: 12.4.0"
  methods: BlobServiceClient.list_containers
  type: bool
include_system:
  description: "Flag specifying that system containers should be included. .. versionadded::
    12.10.0"
  methods: BlobServiceClient.list_containers
  type: bool
incremental_copy:
  description: Copies the snapshot of the source page blob to a destination page blob.
    The snapshot is copied such that only the differential changes between the previously
    copied snapshot are transferred to the destination. The copied snapshots are complete
    copies of the original snapshot and can be read or copied from as usual. Defaults
    to False.
  methods: BlobClient.start_copy_from_url
  type: bool
lease:
  description: If specified, get_container_properties only succeeds if the container"s
    lease is active and matches this ID.
  methods: ContainerClient.get_container_properties | ContainerClient.delete_blob
    | BlobServiceClient.delete_container | BlobClient.get_blob_properties | BlobClient.download_blob
    | BlobClient.upload_blob
  type: Any
legal_hold:
  description: "Specified if a legal hold should be set on the blob. Currently this
    parameter of upload_blob() API is for BlockBlob only.  .. versionadded:: 12.10.0     This
    was introduced in API version ""2020-10-02""."
  methods: BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: bool
length:
  description: Number of bytes to read from the stream. This is optional, but should
    be supplied for optimal performance.
  methods: BlobClient.upload_blob
  type: int
match_condition:
  description: The match condition to use upon the etag.
  methods: ContainerClient.delete_blob | BlobServiceClient.delete_container | BlobClient.get_blob_properties
    | BlobClient.download_blob | BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: azure.core.MatchConditions
max_block_size:
  description: The maximum chunk size for uploading a block blob in chunks. Defaults
    to 4*1024*1024, or 4MB.
  methods: BlobServiceClient
  type: int
max_chunk_get_size:
  description: "The maximum chunk size used for downloading a blob. Defaults to 4*1024*1024,
    or 4MB.  .. admonition:: Example:      .. literalinclude:: ../samples/blob_samples_authentication_async.py         :start-after:
    [START create_blob_service_client]         :end-before: [END create_blob_service_client]         :language:
    python         :dedent: 8         :caption: Creating the BlobServiceClient with
    account url and credential.      .. literalinclude:: ../samples/blob_samples_authentication_async.py         :start-after:
    [START create_blob_service_client_oauth]         :end-before: [END create_blob_service_client_oauth]         :language:
    python         :dedent: 8         :caption: Creating the BlobServiceClient with
    Azure Identity credentials."
  methods: BlobServiceClient
  type: int
max_concurrency:
  description: The number of parallel connections with which to download.
  methods: BlobClient.download_blob | BlobClient.upload_blob
  type: int
max_page_size:
  description: The maximum chunk size for uploading a page blob. Defaults to 4*1024*1024,
    or 4MB.
  methods: BlobServiceClient
  type: int
max_single_get_size:
  description: The maximum size for a blob to be downloaded in a single call, the
    exceeded part will be downloaded in chunks (could be parallel). Defaults to 32*1024*1024,
    or 32MB.
  methods: BlobServiceClient
  type: int
max_single_put_size:
  description: If the blob size is less than or equal max_single_put_size, then the
    blob will be uploaded with only one http PUT request. If the blob size is larger
    than max_single_put_size, the blob will be uploaded in chunks. Defaults to 64*1024*1024,
    or 64MB.
  methods: BlobServiceClient
  type: int
maxsize_condition:
  description: Optional conditional header. The max length in bytes permitted for
    the append blob. If the Append Block operation would cause the blob to exceed
    that limit or if the blob size is already greater than the value specified in
    this header, the request will fail with MaxBlobSizeConditionNotMet error (HTTP
    status code 412 - Precondition Failed).
  methods: BlobClient.upload_blob
  type: int
metadata:
  description: "A dict with name-value pairs to associate with the container as metadata.
    Example: `{""Category"":""test""}`"
  methods: BlobServiceClient.create_container | BlobClient.start_copy_from_url
  type: dict(str, str)
min_large_block_upload_threshold:
  description: The minimum chunk size required to use the memory efficient algorithm
    when uploading a block blob. Defaults to 4*1024*1024+1.
  methods: BlobServiceClient
  type: int
name_starts_with:
  description: Filters the results to return only containers whose names begin with
    the specified prefix.
  methods: BlobServiceClient.list_containers
  type: str
premium_page_blob_tier:
  description: A page blob tier value to set the blob to. The tier correlates to the
    size of the blob and number of allowed IOPS. This is only applicable to page blobs
    on premium storage accounts.
  methods: BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: azure.storage.blob.PremiumPageBlobTier
progress_hook:
  description: "A callback to track the progress of a long running download. The signature
    is function(current: int, total: int) where current is the number of bytes transfered
    so far, and total is the total size of the download."
  methods: BlobClient.download_blob | BlobClient.upload_blob
  type: Any
public_access:
  description: "Possible values include: ""container"", ""blob""."
  methods: BlobServiceClient.create_container
  type: str or azure.storage.blob.PublicAccess
rehydrate_priority:
  description: Indicates the priority with which to rehydrate an archived blob
  methods: BlobClient.start_copy_from_url
  type: azure.storage.blob.RehydratePriority
requires_sync:
  description: Enforces that the service will not return a response until the copy
    is complete.
  methods: BlobClient.start_copy_from_url
  type: bool
results_per_page:
  description: The maximum number of container names to retrieve per API call. If
    the request does not specify the server will return up to 5,000 items.
  methods: BlobServiceClient.list_containers
  type: int
seal_destination_blob:
  description: "Seal the destination append blob. This operation is only for append
    blob. .. versionadded:: 12.4.0"
  methods: BlobClient.start_copy_from_url
  type: bool
secondary_hostname:
  description: The hostname of the secondary endpoint.
  methods: BlobServiceClient
  type: str
source_authorization:
  description: "Authenticate as a service principal using a client secret to access
    a source blob. Ensure "bearer " is the prefix of the source_authorization string.
    This option is only available when `incremental_copy` is set to False and `requires_sync`
    is set to True.  .. versionadded:: 12.9.0"
  methods: BlobClient.start_copy_from_url
  type: str
source_etag:
  description: The source ETag value, or the wildcard character (*). Used to check
    if the resource has changed, and act according to the condition specified by the
    `match_condition` parameter.
  methods: BlobClient.start_copy_from_url
  type: str
source_if_modified_since:
  description: A DateTime value. Azure expects the date value passed in to be UTC.
    If timezone is included, any non-UTC datetimes will be converted to UTC. If a
    date is passed in without timezone info, it is assumed to be UTC. Specify this
    conditional header to copy the blob only if the source blob has been modified
    since the specified date/time.
  methods: BlobClient.start_copy_from_url
  type: datetime.datetime
source_if_unmodified_since:
  description: A DateTime value. Azure expects the date value passed in to be UTC.
    If timezone is included, any non-UTC datetimes will be converted to UTC. If a
    date is passed in without timezone info, it is assumed to be UTC. Specify this
    conditional header to copy the blob only if the source blob has not been modified
    since the specified date/time.
  methods: BlobClient.start_copy_from_url
  type: datetime.datetime
source_lease:
  description: Specify this to perform the Copy Blob operation only if the lease ID
    given matches the active lease ID of the source blob.
  methods: BlobClient.start_copy_from_url
  type: Any
source_match_condition:
  description: The source match condition to use upon the etag.
  methods: BlobClient.start_copy_from_url
  type: azure.core.MatchConditions
standard_blob_tier:
  description: A standard blob tier value to set the blob to. For this version of
    the library, this is only applicable to block blobs on standard storage accounts.
  methods: BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: azure.storage.blob.StandardBlobTier
tags:
  description: "Name-value pairs associated with the blob as tag. Tags are case-sensitive.
    The tag set may contain at most 10 tags.  Tag keys must be between 1 and 128 characters,
    and tag values must be between 0 and 256 characters. Valid tag key and value characters
    include: lowercase and uppercase letters, digits (0-9), space (` `), plus (+),
    minus (-), period (.), solidus (/), colon (:), equals (=), underscore (_)  ..
    versionadded:: 12.4.0"
  methods: BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: Any
timeout:
  description: The timeout parameter is expressed in seconds.
  methods: ContainerClient.get_container_properties | BlobServiceClient.list_containers
    | ContainerClient.walk_blobs | BlobServiceClient.create_container | ContainerClient.delete_blob
    | BlobServiceClient.delete_container | BlobClient.get_blob_properties | BlobClient.download_blob
    | BlobClient.upload_blob | BlobClient.start_copy_from_url
  type: int
use_byte_buffer:
  description: Use a byte buffer for block blob uploads. Defaults to False.
  methods: BlobServiceClient
  type: bool
validate_content:
  description: If true, calculates an MD5 hash for each chunk of the blob. The storage
    service checks the hash of the content that has arrived with the hash that was
    sent. This is primarily valuable for detecting bitflips on the wire if using http
    instead of https, as https (the default), will already validate. Note that this
    MD5 hash is not stored with the blob. Also note that if enabled, the memory-efficient
    upload algorithm will not be used because computing the MD5 hash requires buffering
    entire blocks, and doing so defeats the purpose of the memory-efficient algorithm.
  methods: BlobClient.download_blob | BlobClient.upload_blob
  type: bool
version_id:
  description: "The version id parameter is an opaque DateTime value that, when present,
    specifies the version of the blob to delete.  .. versionadded:: 12.4.0 This keyword
    argument was introduced in API version ""2019-12-12""."
  methods: ContainerClient.delete_blob
  type: str
```

I conclude that passing through kwargs to invocations of azure.storage.blob methods is safe. I will implement this and add well formatted version of this argument list to the readme. 

For the record, here the script used for generating this list:

```python
import azure.storage.blob
import azure.storage.blob.aio
import docstring_parser
import yaml

# checked all invocations from spec.py of azure.storage.blob methods. Explicitly provided arguments to be excluded as second item in tuple.

methods = [
    (azure.storage.blob.aio.BlobServiceClient,['account_url']),
    (azure.storage.blob.ContainerClient.get_container_properties,[]),  # via container exisis
(azure.storage.blob.BlobServiceClient.list_containers,['include_metadata']),  # via ls
(azure.storage.blob.ContainerClient.walk_blobs,['include','name_starts_with']),  # via ls
(azure.storage.blob.ContainerClient.list_blobs,['include','name_starts_with','results_per_page']),  # via ls
(azure.storage.blob.BlobServiceClient.create_container,['name']),  # via mkdir
(azure.storage.blob.ContainerClient.delete_blob,['blob']),  # via rm
(azure.storage.blob.BlobServiceClient.delete_container,['container']),  # via rmdir
(azure.storage.blob.BlobClient.get_blob_properties,['version_id']),  # via isfile
(azure.storage.blob.BlobClient.exists,['version_id']),
(azure.storage.blob.BlobClient.download_blob,['offset','length','version_id']),
(azure.storage.blob.BlobClient.upload_blob,['data','overwrite','metadata','raw_response_hook']),
(azure.storage.blob.BlobClient.start_copy_from_url,['source_url'])]

method_params = list(map(lambda _: (*_,docstring_parser.parse(_[0].__doc__).params),methods))

# check correctly listed explicitly provided argumgents, checked that missing arguments are due to incorrect docstring or docstring parser
for _ in method_params:
    arg_names = [*map(lambda param: param.arg_name,_[-1])]
    for arg_name in _[1]:
        if arg_name not in arg_names:
            print(_[0],arg_name)
        
lookup = {}


for _ in method_params:
    for param in _[-1]:
        if param.arg_name not in _[1]:
            lookup.setdefault(param.arg_name,{'methods': [],'types': [],'defaults':[],'descriptions':[]})
            lookup[param.arg_name]['methods'].append(_[0].__qualname__)
            if param.type_name not in lookup[param.arg_name]['types']:
                lookup[param.arg_name]['types'].append((param.type_name or "Any").replace("~",""))
            if param.default not in lookup[param.arg_name]['defaults']:
                lookup[param.arg_name]['defaults'].append(param.default)
            description = param.description.replace("\n"," ")
            if description not in lookup[param.arg_name]['descriptions']:
                lookup[param.arg_name]['descriptions'].append(description)
            
data = { key: {"methods": " | ".join(value["methods"]),"type": value["types"][0],"description":value["descriptions"][0]} for key,value in lookup.items()}

yaml.dump(data)
```

PR #368, opened 2022-11-17
ls: support listing versions

efiop:
Similar to https://github.com/fsspec/s3fs/pull/661 and https://github.com/fsspec/gcsfs/pull/509


adlfs/spec.py
```diff
@@ -892,6 +901,7 @@ async def _ls(
         delimiter: str = "/",
         return_glob: bool = False,
         version_id: Optional[str] = None,
+        versions: bool = False,
```
efiop, line 904: Turns out `_ls` doesn't support `detail` (even though it states that it does in the docstring). I'll create a pre-requisite PR for that right now (also `_ls` is very crowded and hard to read).

efiop:
@TomAugspurger @TomAugspurger Hey guys, just a friendly ping 🙂 Could you take a look when you'll have time, please?

TomAugspurger:
It would be a while before I can take a look.
________________________________

rlamy:
LGTM. This would be a useful feature.

PR #369, opened 2022-11-17
_ls: support detail

efiop:
The support is stated in the docstring but not implemented in reality.

Also splits `_ls` into `_ls_blobs` and `_ls_containers` (the code itself is intact with an exception of `cache` removed to use `dircache` directly), similar to s3fs and gcsfs, to make it a bit more manageable. I could've slapped `detail` to the bottom of existing `_ls`, but it was way too much code complexity already. Happy to go with that route if the current approach is not well received. 

Pre-requisite for #368

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -521,7 +521,9 @@ class AzureBlobFileSystem(AsyncFileSystem):
         version_id = _coalesce_version_id(path_version_id, kwargs.get("version_id"))
         kwargs["version_id"] = version_id
 
-        out = await self._ls(fullpath, invalidate_cache=invalidate_cache, **kwargs)
+        out = await self._ls(
+            fullpath, detail=True, invalidate_cache=invalidate_cache, **kwargs
+        )
         fullpath = fullpath.rstrip("/")
         out1 = [
             o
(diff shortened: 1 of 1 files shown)


adlfs/spec.py
```diff
@@ -862,32 +870,115 @@ async def _glob(self, path, **kwargs):
...
+                finalblobs = await self._details(
+                    outblobs, target_path=target_path, version_id=version_id
+                )
+                if not finalblobs:
+                    if not await self._exists(target_path):
+                        raise FileNotFoundError
+                    return []
+                if not self.version_aware or finalblobs[0].get("is_current_version"):
+                    self.dircache[target_path] = finalblobs
+                return finalblobs
```
efiop, line 974: This got changed from
```
if self.version_aware or not finalblobs[0].get("is_current_version"):
    return finalblobs
cache[target_path] = finalblobs
self.dircache[target_path] = finalblobs
```

to make it more readable and closer to the way such operations are done elsewhere.

adlfs/spec.py
```diff
@@ -925,109 +1016,22 @@ async def _ls(
         if invalidate_cache:
             self.dircache.clear()
```
efiop, line 1017: We are invalidating cache in one single place here, so we don't need to pass `invalidate_cache` down because `if smth not in self.dircache` will just work.

adlfs/spec.py
```diff
@@ -749,7 +749,10 @@ async def _info(self, path, refresh=False, **kwargs):
         kwargs["version_id"] = version_id
 
         out = await self._ls(
-            self._parent(fullpath), invalidate_cache=invalidate_cache, **kwargs
+            self._parent(fullpath),
+            detail=True,
```
efiop, line 753: Context: we were returning infos by default before, but now `detail` is supported and is `False` by default, so we need to manually ask for detailed results here and in a few other places below.

hayesgb:
Thanks for this PR @efiop .  I like the fact that you broke out `ls_containers` from `ls_blobs`.

I didn't see this before I merged in a change of `info()`, so there are some merge conflicts.  Once those are resolved, I'm good this.

PR #386, opened 2023-01-16
get_file(): don't create local file if the remote doesn't exist

### LINKED ISSUE #385: .get_file() shouldn't create a local file if the remote doesn't exist
When the remote file doesn't exist, `fs.get_file()` raises `FileNotFoundError` (as expected) but nevertheless creates an empty local file.

To reproduce: run `pytest adlfs/tests/test_spec.py -k test_get_file_versioned` and notice that it creates `file.txt` in the cwd.

rlamy:
Fixes #385 and ensures that tests don't litter $PWD.


efiop (approved):
Nice catch! Thank you!

Issue #380, opened 2022-12-13
fs.info() crash when fs is version-aware


rlamy:
When using a version-aware filesystem, `fs.info()` fails on every file (directories are fine).

For instance, the following test fails with a `FileNotFoundError`:
```python
def test_info_version_aware(storage):
    fs = AzureBlobFileSystem(
        account_name=storage.account_name, connection_string=CONN_STR, 
        version_aware=True, skip_instance_cache=True
    )
    file_info = fs.info("data/root/a/file.txt")
```

efiop:
@rlamy Could you post the traceback, please?

PR #390, opened 2023-01-20
Reimplement fs.info() to not depend on fs.ls()

### LINKED ISSUE #377: `fs.info("container")` result depends on dircache status
The result of `fs.info("container")` changes depending on whether `fs.ls("")` was called before or not.

A simple way to reproduce is to run `test_info()` on its own (`pytest adlfs/tests/test_spec.py -k "test_info"`): it fails because there is no `"last_modified"` key in the returned dict.

rlamy:
`fs.info()` now only queries the properties of its argument, instead of potentially doing a full listing of the parent directory, which should improve its performance in many cases.

Fixes #377 and #380.

Note: this PR contains refactorings of related methods that I ended up not modifying otherwise, I could split them off if that makes reviewing easier.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -497,21 +497,6 @@ class AzureBlobFileSystem(AsyncFileSystem):
             version_id if self.version_aware and version_id else None,
         )
 
-    def info(self, path, refresh=False, **kwargs):
-        try:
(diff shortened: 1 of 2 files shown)


hayesgb:
This is nice @rlamy .  Thanks!

PR #392, opened 2023-01-26
Fix missing kwargs injection on upload blob call

nffdiogosilva:
Hello there.

Currently, when uploading a blob, there's no way of injecting other arguments besides the `data`, `overwrite`, and `metadata`.

An example would be when I want to compress data and set the content encoding of the data to `br`. For that I need to pass a `ContentSettings` object into the `upload_blob` method kwargs:

i.e:
```python

from azure.storage.blob import ContentSettings
import brotli
import fsspec

CONN_STRING = "<CONN_STRING>"
CONTAINER = "kwargs-injection"

abfs = fsspec.filesystem(
    "abfs",
    connection_string=CONN_STRING,
)

if __name__ == "__main__":
    content_settings = ContentSettings(
        content_type="application/json",
        content_encoding="br",
    )
    abfs.mkdir(CONTAINER, exist_ok=True)
    abfs.write_bytes(
        path=f"{CONTAINER}/compressed.br",
        value=brotli.compress(b'{"a": "hello world"}', quality=1),
        overwrite=True,
        **{"content_settings": content_settings},
    )

```

What I'm expecting to see in the azure storage containers web interface, when consulting the blob properties:
[](url)
Screenshot 2023-01-26 at 11 22 36

What I see instead (no content encoding defined):
Screenshot 2023-01-26 at 11 35 05

script requirements:
`$ pip install fsspec adlfs brotli`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
     - id: isort
       language_version: python3
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1413,7 +1413,10 @@ class AzureBlobFileSystem(AsyncFileSystem):
             container=container_name, blob=path
         ) as bc:
             result = await bc.upload_blob(
-                data=value, overwrite=overwrite, metadata={"is_directory": "false"}
+                data=value,
+                overwrite=overwrite,
+                metadata={"is_directory": "false"},
+                **kwargs,
             )
         self.invalidate_cache(self._parent(path))
         return result
--- a/docs/index.md
+++ b/docs/index.md
@@ -82,6 +82,14 @@ to list all the files or directories in the top-level of a storage container, yo
 api.md
 ```
 
+**Note**: When uploading a blob (with `write_bytes` or `write_text`) you can injects kwargs directly into `upload_blob` method:
+
+```{code-block} python
+>>> from azure.storage.blob import ContentSettings
+>>> fs = adlfs.AzureBlobFileSystem(account_name="ai4edataeuwest")
+>>> fs.write_bytes(path="path", value=data, overwrite=True, **{"content_settings": ContentSettings(content_type="application/json", content_encoding="br")})
+```
+
 [fsspec]: https://filesystem-spec.readthedocs.io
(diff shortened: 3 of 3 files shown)


TomAugspurger:
xref https://github.com/fsspec/filesystem_spec/issues/916. But agreed that forwarding arguments makes sense, even if that were to be standardized.

We should document this *somewhere* though. Maybe a short note in `index.md`?

nffdiogosilva:
@TomAugspurger that makes sense to standardised but in the meantime can we merge this PR so it can unblock my current use case?

Regarding the note, not sure what are you suggesting. I've added this below the usage section in the index.md:

> **Note**: When uploading a blob (with `write_bytes` or `write_text`) you can injects kwargs directly into `upload_blob` method:
```{code-block} python
>>> from azure.storage.blob import ContentSettings
>>> fs = adlfs.AzureBlobFileSystem(account_name="ai4edataeuwest")
>>> fs.write_bytes(path="path", value=data, overwrite=True, **{"content_settings": ContentSettings(content_type="application/json", content_encoding="br")})
```

Was it something like that you were suggesting?

TomAugspurger (approved):
> that makes sense to standardised but in the meantime can we merge this PR so it can unblock my current use case?

Yep. I think we'll want this PR regardless of what happens upstream, to support other keyword arguments.

It seems that there's a linting check failing. I can't actually view the Azure Pipelines output (cc @hayesgb). Maybe you can try running the linters locally to see what's failing? This looks good otherwise.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
```
nffdiogosilva, line 3: For reference: https://levelup.gitconnected.com/fix-runtimeerror-poetry-isort-5db7c67b60ff

nffdiogosilva:
Hello @TomAugspurger. It seems that the tests were failing due to the isort version that was being used in the pre commit hooks. (you can understand the error better here: https://levelup.gitconnected.com/fix-runtimeerror-poetry-isort-5db7c67b60ff). Fixed that now, and the checks are all passing. Can you merge it, please?

Thank you.

TomAugspurger:
Thanks!

TomAugspurger:
@hayesgb has handled the last few releases. @hayesgb are you able to make a new release sometime, or would you need me to?

eirki:
Just a friendly notification: I would be super grateful if someone could build a new version with this fix.

Issue #348, opened 2022-09-13
Make DefaultAzureCredential as default credential method instead of anonymous access?


QianqianNie:
Hi there

what do you think of making DefaultAzureCredential as the default credential method instead of anonymouns access because I think that is the most common use case.

agrinh:
+1, that would be great.

TomAugspurger:
xref https://github.com/fsspec/adlfs/pull/250, where this was implemented but then reverted. I think we can do this, but we'll need a deprecation cycle.

I was somewhat worried that this would be too noisy of a deprecation, but if we think that the majority of users are specifying `credential` anyway (which I think they are) then the only people affected are those relying on `credential="anon"`.

So +1 to deprecating the current default with a `FutureWarning` and then moving to `DefaultAzureCredential` in a later future.

JoostvDoorn:
Is there any way to move this forward? There are some libraries that do not expose fsspec storage options because they assume the anon=False logic like in s3, and I suppose azure is not their main target audience. As a workaround would it be possible to add an env option to disable anon access before we transition to anon=False by default?

JoostvDoorn:
For those interested in this https://github.com/fsspec/filesystem_spec/pull/1194 provides a generic alternative solution (make sure to update fsspec as it's pretty recent addition). You can pass a default argument to FSSPEC for specific implementations through an env variable as follows:
```python
os.environ["FSSPEC_ABFS"] = """
{
  "anon": false
}
"""
```

I'll be closing my PR (https://github.com/fsspec/adlfs/pull/393) as it's not getting any traction.

PR #393, opened 2023-01-27
Add AZURE_STORAGE_ANON to allow setting anon false

JoostvDoorn:
This will make it easier to configure anon=False for external libraries that do not expose any internal fsspec options.

While a more permanent fix for #348 would be welcomed it would be helpful to have this fix in the meantime.


TomAugspurger:
I'm a little hesitant about inventing new conventions around environment variable names.

Can you share a bit more about the libraries you're using that don't let you configure things? If they're supporting fsspec-style URLs, they should probably expose a `storage_options` argument of some sort, similar to pandas.

JoostvDoorn:
I experience this issue with [pytorch-lightning](https://github.com/Lightning-AI/lightning), and this will partly fix the issues with writing to private blob storage. I was also looking at making [tensorboard](https://github.com/tensorflow/tensorboard) working, as it currently doesn't support azure blob storage, and this may make it one step closer to being compatible.

JoostvDoorn:
Just to be a bit more elaborate on this, using tensorboard cli I can do the following with gs (after `gcloud auth application-default login`):
```
tensorboard --logdir=gs://test-bucket-abcde/some/log/dir
```
With az paths I cannot do that yet.

With this change in this PR and [this small change](https://github.com/JoostvDoorn/tensorboard/commit/128c07007a83a44cb42739efc4bddc50d23a23cd) to tensorboard I would be able to start it with:
```
export AZURE_STORAGE_ACCOUNT_NAME="someaccount"
export AZURE_STORAGE_ANON=false
tensorboard --logdir=az://some-container/some/log/dir
```

PR #396, opened 2023-02-06
Fix ls with protocol prefix

### LINKED ISSUE #395: ls does not work with protocol prefix
Many commands seem to work with a protocol prefix as they call `_strip_protocol` internally however ls doesn't seem to work for ls. It seems consistent with s3fs and gcsfs that this is supported as well by adlfs.

JoostvDoorn:
Fixes #395

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -787,7 +787,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
         """
         logger.debug("abfs.ls() is searching for %s", path)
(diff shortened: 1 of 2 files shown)


JoostvDoorn:
@TomAugspurger any thoughts? Are there any changes necessary? See also https://github.com/fsspec/adlfs/issues/395#issuecomment-1512051707

TomAugspurger:
Looks good, thanks.

PR #398, opened 2023-02-14
Fix recursive upload with directory path

hfurkanvural:
When uploading a directory recursively, the function call hangs forever if it is a directory path. 

It seems there is an issue with async calls but unfortunately I dont know what is it exactly. As the function is inherited from [fsspec](https://github.com/fsspec/filesystem_spec/blob/4143dc16bea64a59efd1ee18ad2b4e560a9b2011/fsspec/spec.py#L288) and it is a noop function, I expect it to return immediately but that is not the case. So, this should fix the issue by immediately returning the function.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1587,7 +1587,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         container_name, path, _ = self.split_path(rpath, delimiter=delimiter)
 
         if os.path.isdir(lpath):
-            return
+            await self._mkdir(rpath)
         else:
             try:
                 with open(lpath, "rb") as f1:
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1587,7 +1587,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         container_name, path, _ = self.split_path(rpath, delimiter=delimiter)
 
         if os.path.isdir(lpath):
-            self.makedirs(rpath, exist_ok=True)
+            return
         else:
             try:
                 with open(lpath, "rb") as f1:


hfurkanvural:
Do you have any comment on this change folks?

@TomAugspurger @hayesgb @efiop

efiop:
@hfurkanvural Is there some reproduction script or something? Sorry, but It is hard to tell what's going on at all right now. 🙁

hfurkanvural:
hey @efiop :)
Do you need any help/assistance here to troubleshoot the issue?

efiop:
@hfurkanvural Sorry, don't have time to debug this. But it looks like you are incorrect about it not being implemented in adlfs. See https://github.com/fsspec/adlfs/blob/2d56e411906519f31ee614e601469f67845b41bd/adlfs/spec.py#L1112 , `_mkdir` is implemented and is not a noop.

efiop (changes requested):
See comment above

hfurkanvural:
@efiop I m afraid you are looking for the wrong function. In the line I wanted to change, the called function is `makedirs()` (which has no implementation) not `_mkdir()`. However there is a function called `makedir()` which is a wrapper around `_mkdir()` thorough a sync_wrapper function. But even though I changed the line and use `makedir()`, the function call never gets a response and get stuck as before and same issue persists.
I believe the behavior here is not correct and causing an issue on the very fundamental functionality.

I ll try to debug more and come up with the actual issue behind.

ghtyrant:
I took some time to further debug this issue and boiled it down to the following MWE:

```python
from fsspec.asyn import sync_wrapper, get_loop


class A:
    def __init__(self):
        self.loop = get_loop()

    @sync_wrapper
    async def fn2(self):
        print("-> fn2()")
        print("<- fn2()")

    @sync_wrapper
    async def fn1(self):
        print("-> fn1()")
        await self.fn2()
        print("<- fn1()")


a = A()
a.fn1()
```
Output:
```
-> fn1()
```
The program never stops and needs to be terminated.

What I encountered during my debugging session is that we run into this piece of code inside the `sync()` function (fsspec/asyn.py):

```python
# (...)
    try:
        loop0 = asyncio.events.get_running_loop()
        if loop0 is loop:
            raise NotImplementedError("Calling sync() from within a running loop")
    except RuntimeError:
        pass
# (...)
```
The `NotImplementedError` is raised when sync() is called the second time (for `fn2`), but is immediately caught and silently ignored. Is that happening intentionally?
After that, the coroutine is scheduled using asyncio and the sync function enters its endless loop to wait for `event` to be set - this never happens.

I appreciate the time you take to look into this @efiop and I hope my findings help you.

ghtyrant:
Another observation that I made: The S3 implementation of fsspec calls `self._mkdir` instead of `self.makedirs`, which I think is the correct call. `self._mkdir` is synchronous, so it doesn't run into the issue described above.

So I'd suggest changing the line from `self.makedirs(rpath, exist_ok=True)` to `await self._mkdir(rpath)`. I tested it and no longer run into a deadlock.

hfurkanvural:
@ghtyrant  thanks for checking it. It seems asyncio doesnt support nested sync() call in a running loop as you mentioned (at least in this context). 
I've updated the line, the function calls directly _mkdir() as @efiop pointed.

MorpheusXAUT:
Is there anything we can do to help you test/verify this and get this merged?
As a tool we're using is dependent on this library in order to work with Azure, it'd be great to have "official" support instead of having to rely on two forked versions :smile:

efiop (approved):

PR #408, opened 2023-03-27
ci: drop azure pipelines

efiop:
No one can even see the test results because it requires some kind of public permissions that we never enabled. Just dropping it and using gha as everyone else.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,56 @@
...
+jobs:
+  tests:
+    timeout-minutes: 45
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]
+        pyv: ["3.8"]
+        reqs: ["earliest", "latest", "dev"]
```
efiop, line 25: Just trying to get feature parity with the old azure-pl-based workflow. Obviously we can test more python versions and more platforms. And also testing 3 reqs is kinda questionable too.

PR #410, opened 2023-04-06
Passing the error message along with FileNotFoundError

KeerthiYandaOS:
If the file is not found, currently adlfs just throws the FileNotFoundError without an error message or the path which is not found.
Having an error message and file path in the error helps in improving the readability of the error.

So in this PR, just updated the spec file to pass the path and message along with the error.


TomAugspurger:
Thanks!

PR #415, opened 2023-05-15
fixes #384

### LINKED ISSUE #384: Container exists / Permissions issue
Hi there

I have a sas_token that somehow does not allow asking for the container properties, only for folders inside. This breaks adlfs's _mkdir which is called by a third-party lib I don't control (pyarrow). Is there some way to not ignore the error on _container_exists if a sas_token was provided?

aersam:
I have to say that I could not make a test because Azurite seems not to support Sas Tokens properly. 

The fix is really simple: Instead of throwing I just print a warning. I'm sure one could do better, but it's a start.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1118,9 +1118,9 @@ class AzureBlobFileSystem(AsyncFileSystem):
             return False
         except Exception as e:
             if self.assume_container_exists is None:
-                logger.warning(
+                warnings.warn(
                     f"Failed to fetch container properties for {container_name}. Assume it exists already",
-                    exc_info=e,
+                    
                 )
                 return True
             else:
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1108,9 +1108,8 @@ class AzureBlobFileSystem(AsyncFileSystem):
         except ResourceNotFoundError:
             return False
         except Exception as e:
-            raise ValueError(
-                f"Failed to fetch container properties for {container_name} for {e}"
-            ) from e
+            logger.warning(f"Failed to fetch container properties for {container_name}. Assume it exists already", exc_info=e)
+            return True
         else:
             return True
 
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -159,7 +159,10 @@ class AzureBlobFileSystem(AsyncFileSystem):
     version_aware : bool (False)
         Whether to support blob versioning.  If enable this will require the
         user to have the necessary permissions for dealing with versioned blobs.
-
+    assume_container_exists: Optional[bool] (None)
+        Set this to true to not check for existance of containers at all, assuming they exist. 
+        None (default) means to warn in case of a failure when checking for existance of a container
+        False throws if retrieving container properties fails
     Pass on to fsspec:
 
     skip_instance_cache:  to control reuse of instances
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks.

I think users should always have some way to silence warnings close to where they occur. Since the user isn't the one calling `_container_exists`, I think our best option is something like a keyword to `AzureBlobFileSystem` like `assume_continer_exists`.

By default, that can be `None`, which is the behavior you've implemented here (try to create it, warn if it doesn't).

Setting `assume_container_exists=True` would just make `_container_exists` a no-op.

Setting `assume_container_exists=False` would restore the previous behavior of trying to create the container and erroring if it doesn't exists.

Would you be able to implement that?

aersam:
Yes I am 🙂

aersam:
> Thanks.
> 
> I think users should always have some way to silence warnings close to where they occur. Since the user isn't the one calling `_container_exists`, I think our best option is something like a keyword to `AzureBlobFileSystem` like `assume_continer_exists`.
> 
> By default, that can be `None`, which is the behavior you've implemented here (try to create it, warn if it doesn't).
> 
> Setting `assume_container_exists=True` would just make `_container_exists` a no-op.
> 
> Setting `assume_container_exists=False` would restore the previous behavior of trying to create the container and erroring if it doesn't exists.
> 
> Would you be able to implement that?

should be done now

adlfs/spec.py
```diff
@@ -159,7 +159,10 @@ class AzureBlobFileSystem(AsyncFileSystem):
     version_aware : bool (False)
         Whether to support blob versioning.  If enable this will require the
         user to have the necessary permissions for dealing with versioned blobs.
-
+    assume_container_exists: Optional[bool] (None)
+        Set this to true to not check for existance of containers at all, assuming they exist. 
+        None (default) means to warn in case of a failure when checking for existance of a container
+        False throws if retrieving container properties fails
```
TomAugspurger, line 165:
```suggestion
        False throws if retrieving container properties fails, which might happen if your
        authentication is only valid at the storage container level, and not the
        storage account level.
```

TomAugspurger (approved):
One small doc update. Otherwise LGTM.

aersam:
Had to push again to make the pre-commit happy, hope it now passes

TomAugspurger:
Looks like a test was relying on the old behavior: https://github.com/fsspec/adlfs/actions/runs/4983934732/jobs/8922081549?pr=415#step:6:114

Maybe set `assume_container_exists=True` in that test (and maybe add tests for the behavior of mkdir with each option? `with pytest.warns` can be used to make sure a warning is emitted?)

aersam:
> Looks like a test was relying on the old behavior: https://github.com/fsspec/adlfs/actions/runs/4983934732/jobs/8922081549?pr=415#step:6:114
> 
> Maybe set `assume_container_exists=True` in that test (and maybe add tests for the behavior of mkdir with each option? `with pytest.warns` can be used to make sure a warning is emitted).

I already did a quick fix with setting this to True. I can add a test for the warning tomorrow, if you want

aersam:
> Looks like a test was relying on the old behavior: https://github.com/fsspec/adlfs/actions/runs/4983934732/jobs/8922081549?pr=415#step:6:114
> 
> Maybe set `assume_container_exists=True` in that test (and maybe add tests for the behavior of mkdir with each option? `with pytest.warns` can be used to make sure a warning is emitted).

Finally added those tests, too

TomAugspurger:
Thanks!

PR #419, opened 2023-06-21
put_file: default to overwrite=True

pmrowla:
adlfs `put_file()` currently raises `FileExistsError` by default if the specified blob already exists. In version aware contexts `put_file` should default to succeeding even if the specified blob already exists, so that put_file creates a new version of the blob.

---

I should also note that as far as I can tell every other fsspec `put_file` implementation defaults to overwriting an existing file, so I'm also wondering whether the current adlfs `overwrite=False` default should be changed in all contexts (regardless of whether or not the fs is version aware)?

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1548,10 +1548,10 @@ class AzureBlobFileSystem(AsyncFileSystem):
         lpath,
         rpath,
         delimiter="/",
-        overwrite=False,
+        overwrite=True,
         callback=None,
         max_concurrency=None,
-        **kwargws,
+        **kwargs,
     ):
         """
         Copy single file to remote
@@ -1559,7 +1559,8 @@ class AzureBlobFileSystem(AsyncFileSystem):
         :param lpath: Path to local file
         :param rpath: Path to remote file
         :param delimitier: Filepath delimiter
-        :param overwrite: Boolean (False).  If True, overwrite the existing file present
+        :param overwrite: Boolean (True). Whether to overwrite any existing file
+            (True) or raise if one already exists (False).
         """
 
         container_name, path, _ = self.split_path(rpath, delimiter=delimiter)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Unreleased
 ----------
 
 - Handle mixed casing for `hdi_isfolder` metadata when determining whether a blob should be treated as a folder.
+- `_put_file`: `overwrite` now defaults to `True`.
 
 2023.10.0
 ---------
--- a/adlfs/tests/test_spec.py
+++ b/adlfs/tests/test_spec.py
@@ -1854,7 +1854,7 @@ async def test_put_file_timeout(storage, mocker, tmp_path):
     upload_blob.assert_called_once_with(
         mocker.ANY,
         metadata={"is_directory": "false"},
-        overwrite=False,
+        overwrite=True,
         raw_response_hook=None,
         max_concurrency=fs.max_concurrency,
         timeout=11,


TomAugspurger:
Thanks for the PR.

I agree that we should match fsspec's default behavior here. I don't think this should depend on whether or not the filesystem is version aware or not.

And it's a bit of a hassle, but I think we should deprecate the current behavior first, by setting the default to `None` which means `overwrite=False` with a warning that it'll change to `True` in the future.

adlfs/spec.py
```diff
@@ -1593,17 +1593,26 @@ async def _expand_path(
...
         :param lpath: Path to local file
         :param rpath: Path to remote file
         :param delimitier: Filepath delimiter
-        :param overwrite: Boolean (False).  If True, overwrite the existing file present
+        :param overwrite: Boolean (True when filesystem is version aware, otherwise
+            False). If True, overwrite the existing file present
         """
 
+        if overwrite is None:
+            overwrite = self.version_aware
```
TomAugspurger, line 1609: Can you explain more about why you think the default behavior should depend on whether the filesystem is version aware? Is it just that you can revert the change in a version-aware filesystem?

IMO, I'd rather have consistent behavior at the library level. I don't think I would want this operation to depend on how the filesystem is configured.

pmrowla: I agree it should be consistent and that it shouldn't depend on the filesystem, but I don't think the existing `overwrite=False` default makes sense to have at all for the version-aware use case.

So if the preference is to keep the existing default for non-versioned filesystems in the short term, I think there should be separate behavior for the version-aware use case until the future change is made to make the default `overwrite=True` in all case.

TomAugspurger: Can you check the other fsspec implementations? I believe that s3 / s3fs should have versioned filesystems. I'd prefer to match that behavior.

pmrowla: s3fs and gcsfs don't have the `overwrite` parameter, and both implementations always overwrite on `put_file()` regardless of whether or not the fs is versioned

efiop: I agree that making `overwrite` default value dependant on `version_aware` feels really odd. Feels like for the sake of consistency it is better to switch overwrite default to True and then we won't need to do anything else here, right?

TomAugspurger: Just seems considerate to people who are relying on the current behavior.

But... now that I think of it, users of adlfs through higher-level systems like dask and pandas won't actually have a chance to set `overwrite=True`. It would only be helpful if you're using it directly (or if we jumped through a bunch of hoops and made it a setting on the filesystem too, which sounds too complicated).

So I guess I'm OK with just changing the default to match other the libraries.

adlfs/spec.py
```diff
@@ -1593,17 +1593,26 @@ async def _expand_path(
         return list(sorted(out))
 
     async def _put_file(
-        self, lpath, rpath, delimiter="/", overwrite=False, callback=None, **kwargws
+        self, lpath, rpath, delimiter="/", overwrite=None, callback=None, **kwargs
```
TomAugspurger, line 1596:
```suggestion
        self, lpath, rpath, delimiter="/", overwrite=True, callback=None, **kwargs
```

adlfs/spec.py
```diff
@@ -1593,17 +1593,26 @@ async def _expand_path(
...
     ):
         """
         Copy single file to remote
 
         :param lpath: Path to local file
         :param rpath: Path to remote file
         :param delimitier: Filepath delimiter
-        :param overwrite: Boolean (False).  If True, overwrite the existing file present
+        :param overwrite: Boolean (True when filesystem is version aware, otherwise
+            False). If True, overwrite the existing file present
```
TomAugspurger, line 1605:
```suggestion
        :param overwrite: Boolean (True). Whether to overwrite any
        existing file (True) or raise if one already exists (False).
```

adlfs/spec.py
TomAugspurger, line 1609:
```suggestion
```

adlfs/spec.py
```diff
@@ -1593,17 +1593,26 @@ async def _expand_path(
...
         """
 
+        if overwrite is None:
+            overwrite = self.version_aware
+            if not overwrite:
+                warnings.warn(
+                    "overwrite defaulting to False has been deprecated. The default"
+                    " behavior will be changed to overwrite=True in future releases.",
+                    FutureWarning,
+                )
```
TomAugspurger, line 1615:
```suggestion
```

TomAugspurger (commented):
We can clean this up if we're just changing stuff to `overwrite=True`.

We'll need a note in the changelog. Maybe a versionchanged in the docstring too.

pmrowla:
Updated the PR as requested

efiop (approved):

Issue #268, opened 2021-08-19
parallel download support


lmeyerov:
In benchmarks, we're finding significant speedups by using parallel downloads for MB+ individual blobs:

for a 200MB file on an az server <> az store, warm:

- fs.download: 4.9s
- file.write(await (await bc.download_blob()).readall())): 4.3s
- file.write(await (await bc.download_blob(max_concurrency=16)).readall())): 1.3s
- trickier: I think it's even bigger in practice b/c above was a warm read. cold read baselines were more 10-20s.

This is on a node with 8 cores and 2 NICs, and I think < 1 Gbps

Ideally:
- smart default: concurrency defaults to # cores x2 or x4, something like that
- unclear: handling of recursive case
- allow manual overrides
- works with dask, e.g., dd.read_parquet(...)

hayesgb:
@lmeyerov -- Really appreciate this benchmarking!  I've put together a crack at this.  It's currently implemented in the `concurrent_io` branch for:

```
fs.get_file()
fs.put_file()
fs.open()
AzureBlobFile.write()
```

When I timed this on my local machine, `put_file` gave me over 50% reduction in write times, so I'd appreciate any feedback you have.

PR #420, opened 2023-08-03
support `max_concurrency` in `upload_blob` and `download_blob` operations

pmrowla:
- Exposes `max_concurrency=None` kwarg in async fs methods that use the azure SDK `upload_blob` and `download_blob` (which azure then uses to parallelize async chunk uploads/downloads)
- Adds fs level `AzureBlobFileSystem.max_concurrency` attribute which is used whenever method-level `max_concurrency` is not set
    - Defaults to `fsspec.asyn._get_batch_size()`
- For fs methods that support multiple file batching, `max_concurrency` defaults to 1 for the individual file upload/download.
    - So by default `batch_size=...` is used to parallelize uploads/downloads at the file level (in something like `fs._get()` or `fs._put()`), and no additional parallelization is done for chunks within each file.
    - To get both file level and chunk level parallelization the user can set both `batch_size` and `max_concurrency`. i.e. `fs.get(path, batch_size=4, max_concurrency=2)` would download up to 4 files at a time, and up to 2 chunks at a time within each file, giving an overall concurrency of up to 8 async download coroutines being run in the loop at a time.

Closes https://github.com/fsspec/adlfs/issues/268 (and supercedes the changes in the `concurrent_io` branch)

This PR is incompatible with https://github.com/fsspec/adlfs/pull/329 (but there was discussion in that PR regarding changing the name of the parameter used to something other than `max_concurrency` since it conflicts with the the azure SDK parameter)

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -297,7 +297,11 @@ class AzureBlobFileSystem(AsyncFileSystem):
         if self.credential is not None:
             weakref.finalize(self, sync, self.loop, close_credential, self)
 
-        self.max_concurrency = max_concurrency or (_get_batch_size() // 4)
+        if max_concurrency is None:
+            batch_size = _get_batch_size()
+            if batch_size > 0:
+                max_concurrency = batch_size
+        self.max_concurrency = max_concurrency
 
     @classmethod
     def _strip_protocol(cls, path: str):
@@ -1453,6 +1457,12 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
     pipe_file = sync_wrapper(_pipe_file)
 
+    async def _pipe(self, *args, batch_size=None, max_concurrency=None, **kwargs):
+        max_concurrency = max_concurrency or 1
+        return await super()._pipe(
+            *args, batch_size=batch_size, max_concurrency=max_concurrency, **kwargs
+        )
+
     async def _cat_file(
         self, path, start=None, end=None, max_concurrency=None, **kwargs
     ):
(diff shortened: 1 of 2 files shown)


adlfs/tests/test_spec.py
```diff
@@ -392,7 +392,9 @@ def test_info_missing(storage, path):
 
 def test_time_info(storage):
     fs = AzureBlobFileSystem(
-        account_name=storage.account_name, connection_string=CONN_STR
+        account_name=storage.account_name,
+        connection_string=CONN_STR,
+        max_concurrency=1,
```
pmrowla, line 397: Concurrent upload cannot be used for this test, otherwise `storage.insert_time` will be the timestamp of the first completed chunk and `creation_time` will be the timestamp of the finished operation (after all chunks are uploaded).

efiop (approved):

efiop:
@hayesgb @TomAugspurger Hey folks. Could you take a look, please? Just want to make sure you are fine with this.

adlfs/spec.py
```diff
@@ -26,7 +26,7 @@
 from azure.storage.blob._shared.base_client import create_configuration
 from azure.storage.blob.aio import BlobServiceClient as AIOBlobServiceClient
 from azure.storage.blob.aio._list_blobs_helper import BlobPrefix
-from fsspec.asyn import AsyncFileSystem, get_loop, sync, sync_wrapper
+from fsspec.asyn import AsyncFileSystem, _get_batch_size, get_loop, sync, sync_wrapper
```
TomAugspurger, line 29: Can we work in fsspec to expose a public version of this? Or at least get a promise that this isn’t going to change and break stuff here?

TomAugspurger (approved):
Looks good. Just one question but feel free to merge.

adlfs/spec.py
pmrowla: https://github.com/fsspec/filesystem_spec/issues/1327

efiop:
Btw, for the record, some comparisons to show how much this patch speeds things up for us https://github.com/iterative/dvc-azure/issues/54#issuecomment-1663636862

PR #422, opened 2023-09-07
Fix glob behaviour

ap--:
Hello everyone,

I noticed that `fsspec==2023.9.0` fixed the way `.glob()` handles "**" patterns. fsspec/filesystem_spec#1329
This PR updates `adlfs` to rely on fsspec's newer glob implementation.

I opted for bumping the minimum required fsspec version, but this could also be handled by vendoring the newer fsspec glob implementations for older fsspec versions.

Let me know if you'd prefer the other, and I can change the PR.

Cheers,
Andreas 😃


TomAugspurger:
Thanks for the PR. I agree with the decision to bump the minimum version of fsspec.

Looking at https://github.com/fsspec/filesystem_spec/pull/1329, it seems there are quite a few issues linking to it, so this is a disruptive change. But maintaining compatibility with fsspec is the most important thing here. Hopefully we don't ever have to change the behavior of glob again.

I'll merge once this is green.

TomAugspurger (approved):

TomAugspurger:
Thanks @ap--! I'll look into a getting a release out soon.

Issue #424, opened 2023-09-13
"sdk_moniker" key error 


benrutter:
This is an error that's coming downstream from aszure-sdk-for-python since yesterdays (azure-storage-blob) release (12.8.0)

It happens at initialisation of AzureBlobFileSystem() if azure-storage-blob version is 12.8.0. Here's the traceback:

```python
file /databricks/python/lib/python3.10/site-packages/adlfs/spec.py:222, in AzureBlobFileSystem()
    211 protocol = "abfs"
    213 def __init__(
    214     self,
    215     account_name: str = None,
    216     account_key: str = None,
    217     connection_string: str = None,
    218     credential: str = None,
    219     sas_token: str = None,
    220     request_session=None,
    221     socket_timeout=_SOCKET_TIMEOUT_DEFAULT,
--> 222     blocksize: int = create_configuration(storage_sdk="blob").max_block_size,
    223     client_id: str = None,
    224     client_secret: str = None,
    225     tenant_id: str = None,
    226     anon: bool = True,
    227     location_mode: str = "primary",
    228     loop=None,
    229     asynchronous: bool = False,
    230     default_fill_cache: bool = True,
    231     default_cache_type: str = "bytes",
    232     version_aware: bool = False,
    233     assume_container_exists: Optional[bool] = None,
    234     max_concurrency: Optional[int] = None,
    235     **kwargs,
    236 ):
    237     super_kwargs = {
    238         k: kwargs.pop(k)
    239         for k in ["use_listings_cache", "listings_expiry_time", "max_paths"]
    240         if k in kwargs
    241     }  # pass on to fsspec superclass

File /databricks/python/lib/python3.10/site-packages/azure/storage/blob/_shared/base_client.py:415, in create_configuration(**kwargs)
    414 config.headers_policy = StorageHeadersPolicy(**kwargs)
--> 415 config.user_agent_policy = UserAgentPolicy(sdk_moniker=kwargs.pop('sdk_moniker'), **kwargs)
    416 config.retry_policy = kwargs.get("retry_policy") or ExponentialRetry(**kwargs)

KeyError: 'sdk_moniker'
```

Seems like the issue is this line:
```python
config.user_agent_policy = UserAgentPolicy(sdk_moniker=kwargs.pop('sdk_moniker'), **kwargs)
```
which effictively makes "sdk_moniker" a required keyword argument.

[...]

PR #425, opened 2023-09-13
azure sdk_moniker kwarg bug (downstream from azure-storage-blob)

benrutter:
Not sure if this is the appropriate fix or not? This PR would introduce compatibilty with azure-storage-blob version 12.18.0 which now requires sdk_moniker as a keyword argument.

For further details, [see issue 424](https://github.com/fsspec/adlfs/issues/424)

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -219,7 +219,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         sas_token: str = None,
         request_session=None,
         socket_timeout=_SOCKET_TIMEOUT_DEFAULT,
-        blocksize: int = create_configuration(storage_sdk="blob").max_block_size,
+        blocksize: int = create_configuration(storage_sdk="blob", sdk_moniker="storage-blob").max_block_size,
         client_id: str = None,
         client_secret: str = None,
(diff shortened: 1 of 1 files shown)


benrutter:
Ah, this won't actually work, sorry!

azure-storage-blob pre 12.18.0 looks like this:
```python
config.user_agent_policy = UserAgentPolicy(
    sdk_moniker=(f"storage-{kwargs.pop('storage_skd')}/{VERSION}", **kwargs)
)
```
and, it *now* look like this:
```python
config.user_agent_policy = UserAgentPolicy(sdk_moniker=kwargs.pop('sdk_moniker'), **kwargs)
```
Effectively, if you pass in sdk_moniker in a version before 12.18.0 you'll get both a dynamically generated skd_moniker, and the one you've specified.

If you *don't* pass it in, for version 12.18.0, you'll get an error.

So aside from doing something like figuring out which version of azure-store-blob is installed and then acting conditionally on that, I don't think it's possible to use create_configuration in a way that's consistent across versions.

If they haven't already spotted this, I'll let the azure-blob-storage project know this - I'm assuming its an accident.

Issue #426, opened 2023-09-13
Avoid private APIs from azure.storage


TomAugspurger:
- [ ] https://github.com/fsspec/adlfs/blob/4b704b1a5f39ec2350a8be44d9cb117aee9b1719/adlfs/spec.py#L25
- [ ] https://github.com/fsspec/adlfs/blob/4b704b1a5f39ec2350a8be44d9cb117aee9b1719/adlfs/spec.py#L26
- [ ] https://github.com/fsspec/adlfs/blob/4b704b1a5f39ec2350a8be44d9cb117aee9b1719/adlfs/spec.py#L28

jalauzon-msft:
Just to help out a little, `BlobBlock`, `BlobProperties`, `BlobType`, and `BlobPrefix` are all exported from the main SDK so should all be able to be imported directly from `azure-storage-blob`.

`create_configuration` is not exported and is not really recommended to be used, especially just to get the default block size. Unfortunately, we (the SDK team) don't really have an export for our default block size at this moment, but it is 4 MiB (4 * 1024 * 1024) and not likely to change anytime soon so you could consider just hardcoding it. If you really think you want that constant exported, feel free to open an issue to the SDK and we can consider adding it (but may ultimately decide not to).

benrutter:
create_configuration probably needs a bit of discussion, but I'd be down to put in a PR for this to help get it fixed. It should just be a drop in replacement on the imports and then a quick check that no tests are broken.

PR #427, opened 2023-09-14
Avoid use of private APIs from azure.storage

benrutter:
This is designed to fix [issue 426](https://github.com/fsspec/adlfs/issues/426), and remove use of private modules from azure.storage - hopefully help avoid future issues like the recent [sdk_moniker key error](https://github.com/fsspec/adlfs/issues/424) cropping in.

There's two changes here, both of which are pretty minor, but the second of which might need some discussion on whether it's the right approach.

## Import switches

```BlobPrefix```, ```BlobBlock```, ```BlobPoperties``` and ```BlobType ```are all exposed as importables straight from azure.storage.blob without needing the private modules, so this is a drop in replacement.

## Default Block Size

This change might need a little more discussion - I've hardcoded it here to avoid using ```create_configuration```, which @jalauzon-msft suggested as a possibility, but that means it won't automatically change if azure shifts it in future.

The alternative approaches I can think of are:
- Continue using ```create_configuration```, which would be responsive to changes from azure, but also potentially lead to issues in future since it is a private method.
- Open an issue with azure SDK to add default block size as a supported importable constant (and either leave as is or hardcode in the meantime)

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -23,12 +23,12 @@ from azure.core.exceptions import (
 )
 from azure.storage.blob import (
     BlobBlock,
-    BlobPrefix,
     BlobProperties,
     BlobSasPermissions,
     BlobType,
     generate_blob_sas,
 )
+from azure.storage.blob.aio import BlobPrefix
 from azure.storage.blob.aio import BlobServiceClient as AIOBlobServiceClient
 from fsspec.asyn import AsyncFileSystem, _get_batch_size, get_loop, sync, sync_wrapper
 from fsspec.spec import AbstractBufferedFile
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -21,11 +21,15 @@ from azure.core.exceptions import (
     ResourceExistsError,
     ResourceNotFoundError,
 )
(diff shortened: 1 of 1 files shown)


benrutter:
Looks like I did something unexpected since all the tests are now failing - not sure I understand where the errors are coming from though? (I've checked, and the new imports are definitely just extended ways of pulling from the initially private modules, so I don't think I've accidently switched out an azure object for another one)

TomAugspurger:
Thanks @benrutter! Agreed with hardcoding the default block size and getting that from a public API if and when it's added to the azure SDK.

I wonder if the failing tests are fixed by https://github.com/fsspec/adlfs/pull/422. I'll take a closer look later in the week at these two PRs.

TomAugspurger:
Hmm, seems that I can't push to your branch @benrutter. Are you able to merge main and update this PR?

TomAugspurger:
https://github.com/fsspec/adlfs/pull/427/commits/78af9d177bea2408af514475bbbf67fd14fe19d6 should fix the CI failure (sync vs. async BlobPrefix)

TomAugspurger:
Thanks!

benrutter:
Thanks @TomAugspurger - just saw your messages now!

PR #437, opened 2023-11-24
Make AzureBlobFileSystem anon behaviour configurable via env var.

microft:
Making the `anon` default value configurable from an env var makes integration easier with helm charts that provide features like RBAC in Azure accounts.

In a multi-cloud application I am working on it is impacting that the code needs to change between AWS and Azure. 
The default in s3fs is anon=False, in adlfs anon=True. 

This PR would allow for configuration only to change without any code change. 

It would facilitate integration with https://azure.github.io/azure-workload-identity/docs/quick-start.html without any code change apart from environment variables.


adlfs/spec.py
```diff
@@ -268,7 +268,7 @@ def __init__(
         self.client_id = client_id or os.getenv("AZURE_STORAGE_CLIENT_ID")
         self.client_secret = client_secret or os.getenv("AZURE_STORAGE_CLIENT_SECRET")
         self.tenant_id = tenant_id or os.getenv("AZURE_STORAGE_TENANT_ID")
-        self.anon = anon
+        self.anon = anon or os.getenv("AZURE_STORAGE_ANON", True)
```
TomAugspurger, line 271: What kind of values are you expecting here? I think you'll need to cast from whatever `os.getenv("AZURE_STORAGE_ANON")` is to some kind of bool. I think we want something like `"0"` to be `False`. And maybe `"false"` to be `False` as well.

We'll need to document this.

TomAugspurger (commented):
Thanks for this PR. A few more requests:

1. Add docs for this to the class docstring
2. Add docs in the `README`
3. Add a test

adlfs/spec.py
TomAugspurger: Sorry, I had it backwards :/ Your proposal looks good.

TomAugspurger:
Are you able to add a test and docs?

I think pytest has a fixture to monkeypatch environment variables for a test. Just need to confirm that it gets set in `__init__` properly.

TomAugspurger:
Checking in here, will you have a chance to update these last things @microft?

microft:
Will try to update in the next few days. 
(Festivities plus child birth have kept me mostly offline)

microft:
@TomAugspurger I got a test going, but the `cachable` property (inherited from AbstractFileSystem) seems to complicate testing multiple instances with the same `account_name` and `connection_string`. Hence the monkey patch on `cachable` for that test. 
Hope it makes sense.

r-priyam:
@TomAugspurger: Any idea when this can be approved and merged, please? It's affecting us badly, too.

PR #441, opened 2023-12-13
callback: guard against invalid values

skshetry:
See https://github.com/Azure/azure-sdk-for-python/issues/11419#issuecomment-628143480 on how to get progress information.
Also related: https://github.com/fsspec/adlfs/pull/275#issuecomment-1119411234 and https://github.com/iterative/dvc/issues/7704.

Invalid values should not be passed to callbacks.

cc @efiop

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -82,7 +82,9 @@ def make_callback(key, callback):
 
         current = response.context.get(key)
         total = response.context["data_stream_total"]
-        if not sent_total:
+        if current is None:
+            return
+        if not sent_total and total is not None:
             callback.set_size(total)
         callback.absolute_update(current)
 


.github/workflows/docs.yml
```diff
@@ -11,7 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/setup-python@v5
+      with:
+        python-version: 3.11
```
skshetry, line 16: Without this, it'll use Python 3.12. `versioneer.py` file needs to be updated for that to work with 3.12.

skshetry:
Tests are failing, but seems unrelated. @efiop, could you please take a look?

adlfs/spec.py
```diff
@@ -82,7 +82,9 @@ def wrapper(response):
 
         current = response.context.get(key)
         total = response.context["data_stream_total"]
-        if not sent_total:
+        if current is None:
+            return
+        if not sent_total and total is not None:
```
skshetry, line 87: `sent_total` is not even being used here.

Should probably be set inside to `True`, but 🤷🏽.

efiop:
@skshetry Fixed CI, please rebase.

.github/workflows/docs.yml
efiop: @skshetry Dropped versioneer, please rebase and you will be able to safely remove this.

efiop (approved):

PR #444, opened 2023-12-22
adlfs: raise nicer FileNotFoundError's

efiop:
Noticed that they were all over the place and often didn't even have the path set, so decided to fix what I could see.


efiop:
Tests are broken because of changes in https://github.com/fsspec/filesystem_spec/pull/1382 , will handle in a followup.

PR #445, opened 2023-12-22
gha: run tests on schedule

efiop:
Just to stay on top of dependencies and see if something broke.


efiop:
Looks like versioneer is acting up, time to throw it out and replace with setuptools_scm, will do in a followup.

PR #448, opened 2023-12-22
adlfs: don't append / to dirs in find() and fix glob()

efiop:
This makes it consistent with localfs/gcsfs/s3fs and also makes glob work after changes in https://github.com/fsspec/filesystem_spec/pull/1382

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -880,12 +880,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
         return output
 
-    def find(self, path, withdirs=False, prefix="", **kwargs):
-        return sync(
-            self.loop, self._find, path=path, withdirs=withdirs, prefix=prefix, **kwargs
-        )
-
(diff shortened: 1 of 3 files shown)


adlfs/spec.py
```diff
@@ -880,12 +880,7 @@ async def _details(
 
         return output
 
-    def find(self, path, withdirs=False, prefix="", **kwargs):
-        return sync(
-            self.loop, self._find, path=path, withdirs=withdirs, prefix=prefix, **kwargs
-        )
```
efiop, line 886: This was redundant.

adlfs/spec.py
```diff
@@ -960,8 +952,6 @@ async def _find(self, path, withdirs=False, prefix="", with_parent=False, **kwar
                 files[file["name"]] = file
 
         if withdirs:
-            if not with_parent:
-                dirs.pop(target_path, None)
```
efiop, line 964: with_parent is not present in fsspec/gcsfs/s3fs nor it is explicitly tested anywhere. It was getting in the way, so I've removed it as well.

PR #451, opened 2024-01-03
fs.url(): expose response content headers for pre-signed URLs

pmrowla:
Exposes the fields for setting Content-Disposition, Content-Type, Content-Encoding, Content-Language response headers on a pre-signed URL. This can be useful when the desired user-facing downloaded filename is not necessarily the same as the blob path name.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1454,7 +1454,16 @@ class AzureBlobFileSystem(AsyncFileSystem):
     def url(self, path, expires=3600, **kwargs):
         return sync(self.loop, self._url, path, expires, **kwargs)
 
-    async def _url(self, path, expires=3600, **kwargs):
+    async def _url(
+        self,
+        path,
+        expires=3600,
+        content_disposition=None,
+        content_encoding=None,
+        content_language=None,
+        content_type=None,
+        **kwargs,
+    ):
         """Generate presigned URL to access path by HTTP
 
         Parameters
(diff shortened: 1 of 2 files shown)


adlfs/spec.py
```diff
@@ -1637,6 +1658,10 @@ def download(self, rpath, lpath, recursive=False, **kwargs):
         """Alias of :ref:`FilesystemSpec.get`."""
         return self.get(rpath, lpath, recursive=recursive, **kwargs)
 
+    def sign(self, path, expiration=100, **kwargs):
+        """Create a signed URL representing the given path."""
+        return self.url(path, expires=expiration, **kwargs)
```
pmrowla, line 1663: s3fs uses `url()` for generating signed URLs, which I'm assuming is why this was originally implemented as `url()` in adlfs, but the top level fsspec defines actually defines `sign()` for this purpose:
https://github.com/fsspec/filesystem_spec/blob/ac589d8318d761ab6e198e00ba6cbd6c663732d7/fsspec/spec.py#L1536C5-L1537C59

(`s3fs.sign()` just calls `s3fs.url()`)

adlfs/spec.py
```diff
@@ -1474,6 +1491,10 @@ async def _url(self, path, expires=3600, **kwargs):
             permission=BlobSasPermissions(read=True),
             expiry=datetime.utcnow() + timedelta(seconds=expires),
             version_id=version_id,
+            content_disposition=content_disposition,
+            content_encoding=content_encoding,
+            content_language=content_language,
+            content_type=content_type,
```
pmrowla, line 1497: In s3fs this is supported by passing `**kwargs` directly into the aws/boto call, but it seems safer to be more explicit rather than passing kwargs into the `generate_blob_sas`.

pmrowla: `gcsfs.sign()` also just passes `**kwargs` directly into the signing call, so I would not be opposed to making that change here if it's preferred

efiop (approved):

Issue #389, opened 2023-01-20
Recursive remove do not work with newest versions


PeterFogh:
Hi, my old conda environment:
```
adlfs                     2022.11.2          pyhd8ed1ab_0    conda-forge
azure-core                1.26.1             pyhd8ed1ab_0    conda-forge
azure-datalake-store      0.0.51             pyh9f0ad1d_0    conda-forge
azure-identity            1.12.0             pyhd8ed1ab_0    conda-forge
azure-storage-blob        12.14.1            pyhd8ed1ab_0    conda-forge
fsspec                    2022.10.0       py310haa95532_0
```
runs the following without any error
```
import adlfs
import azure.identity.aio
AZURE_FS = adlfs.AzureBlobFileSystem(
    account_name=STORAGE_ACCOUNT,
    credential=azure.identity.aio.DefaultAzureCredential())
AZURE_FS.rm(PATH, recursive=True)
```
But after updating to these package versions:
```
adlfs                     2023.1.0           pyhd8ed1ab_0    conda-forge
azure-core                1.26.2             pyhd8ed1ab_0    conda-forge
azure-datalake-store      0.0.51             pyh9f0ad1d_0    conda-forge
azure-identity            1.12.0             pyhd8ed1ab_0    conda-forge
azure-storage-blob        12.13.1            pyhd8ed1ab_0    conda-forge
fsspec                    2022.11.0       py310haa95532_0
```
it raises this error:
```RuntimeError: ('Failed to remove %s for %s', [PATHS], ResourceExistsError('This operation is not permitted on a non-empty directory.\nRequestId:ID\nTime:2023-01-20T09:33:26.1693752Z\nErrorCode:DirectoryIsNotEmpty'))```

Bacially, the `recursive=True` do not work as intended.

efiop:
CC @daavoo in case it is related to https://github.com/fsspec/adlfs/pull/383

daavoo:
@PeterFogh , what is `PATH` in the snippet?

PeterFogh:
I have tried both direct paths like "container/blob" and protocal paths link "az://container/blob" - but both fail the same way.

daavoo:
Hi @PeterFogh , I am unable to reproduce. Does the dir have any particular structure?


[...]

PR #383, opened 2023-01-10
rm: Optimizations for multiple files.

daavoo:
- Don't call `_isfile`.
- Group files by `container_name` and use `asyncio.gather` to remove the groups.
- Allow to skip `_expand_path`.

Little benchmark I ran:

branch | time
----- | ---------------------
main | 51.00587606430054
PR `expand_path=True` (default) | 20.936074018478394
PR `expand_path=False` | 1.5207679271697998

PR #453, opened 2024-01-24
Revert #383

Tom-Newton:
Resolves https://github.com/fsspec/adlfs/issues/389

Unfortunately https://github.com/fsspec/adlfs/pull/383 broke recursive delete on blob storage accounts with hierarchical namespace enabled (ADLS gen2). 

```
azure.core.exceptions.ResourceExistsError: This operation is not permitted on a non-empty directory.
RequestId:f3042353-701e-0069-7a68-dc7b41000000
Time:2023-09-01T00:08:11.0327332Z
ErrorCode:DirectoryIsNotEmpty
Content: <?xml version="1.0" encoding="utf-8"?><Error><Code>DirectoryIsNotEmpty</Code><Message>This operation is not permitted on a non-empty directory.
RequestId:f3042353-701e-0069-7a68-dc7b41000000
Time:2023-09-01T00:08:11.0327332Z</Message></Error>
```

This is because https://github.com/fsspec/adlfs/pull/383 causes all blobs (including hierarchical namespace directory markers) do be deleted asynchronously. There is nothing to ensure that the contents of a directory are deleted before the directory itself, hence the error above. 

There are some proposals about how to fix this without reverting discussed on the issue https://github.com/fsspec/adlfs/issues/389#issuecomment-1845899468. However those will require some thinking and testing. If we want to fix it properly I think the best way would be to have separate code paths for hierarchical vs flat namespace but that would be significantly more complicated (see https://github.com/apache/arrow/blob/2e8bd8d0b53560a561656337021abc3e4aa73f8c/cpp/src/arrow/filesystem/azurefs.cc#L1822-L1848 for an example). I think its most important to just get back to a state where it works.  

I hope this doesn't come across as rude to revert someone else's contribution without their acknowledgement. I did [try to gain some consensus](https://github.com/fsspec/adlfs/issues/389#issuecomment-1866816269) and nobody responded for over a month. Being stuck on `2022.11.0` is causing us a lot of issues.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -11,7 +11,6 @@ import os
 import re
 import warnings
 import weakref
-from collections import defaultdict
 from datetime import datetime, timedelta
 from glob import has_magic
 from typing import Optional, Tuple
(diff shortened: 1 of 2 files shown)


Tom-Newton:
@efiop please could you review? Sorry for the direct ping, but you are the only maintainer I can see on https://github.com/fsspec/adlfs/issues/389.

efiop:
@Tom-Newton No worries. Seems like you worked on a proper solution in https://github.com/fsspec/adlfs/issues/389#issuecomment-1845899468 , what's the status of it?

efiop:
If directory deletion is the only problem, there should be a straightforward fix for it (you did sorting which seems to work). Two logic branches shouldn't be a problem either. Considering how deep you are already, it does seem a bit wasteful to just revert and burry the optimization. Maybe that's the best option now, but I'm just trying to explore.

Tom-Newton:
I think https://github.com/Tom-Newton/adlfs/pull/1 works and does not revert the optimisation. There is a bit more risk with that and I have not done extensive testing. Its possible I missed something and https://github.com/Tom-Newton/adlfs/pull/1 just makes the problem a lot less common. 

Maybe I can find some time to write an extra test for https://github.com/Tom-Newton/adlfs/pull/1 (it won't be a good test because azurite doesn't support hierarchical namespace).

Tom-Newton:
Closing in favour of https://github.com/fsspec/adlfs/pull/454 which fixes forward rather than reverts.

PR #454, opened 2024-01-26
Fix recursive delete on hierarchical namespace accounts

Tom-Newton:
Resolves: https://github.com/fsspec/adlfs/issues/389

Alternative fix to https://github.com/fsspec/adlfs/pull/453

Unfortunately https://github.com/fsspec/adlfs/pull/383 broke recursive delete on blob storage accounts with hierarchical namespace enabled (ADLS gen2). 

```
azure.core.exceptions.ResourceExistsError: This operation is not permitted on a non-empty directory.
RequestId:f3042353-701e-0069-7a68-dc7b41000000
Time:2023-09-01T00:08:11.0327332Z
ErrorCode:DirectoryIsNotEmpty
Content: <?xml version="1.0" encoding="utf-8"?><Error><Code>DirectoryIsNotEmpty</Code><Message>This operation is not permitted on a non-empty directory.
RequestId:f3042353-701e-0069-7a68-dc7b41000000
Time:2023-09-01T00:08:11.0327332Z</Message></Error>
```

This is because https://github.com/fsspec/adlfs/pull/383 causes all blobs (including hierarchical namespace directory markers) do be deleted asynchronously. There is nothing to ensure that the contents of a directory are deleted before the directory itself, hence the error above. 

### Chosen fix
Look at the sorted list of blob names to derive which blobs are directory markers so they can be deleted in the correct order. Other blobs are deleted asynchronously. The strategy used for this is a bit strange but I think its the best option, at least for now. 

### Alternative solutions:
1. `container_client.delete_blobs()`
    1. A single call to Azure makes a batch delete. Not sure how the performance will compare to the async single deleted that we do. 
    2. This looks like its works on hierarchical and flat namespaces storage accounts but it doesn't work on `azurite` https://github.com/Azure/Azurite/issues/1809
    3. Simple solution if we have a list of blobs to delete. 
1. Detect a hierarchical namespace accounts and follow a different code path
    1.  Detecting hierarchical namespace is shockingly difficult to do e.g. https://github.com/apache/arrow/blob/667e9170ef363e9b4a067be3be245d2dcd4b7a6f/cpp/src/arrow/filesystem/azurefs.cc#L832-L879
    1. The `delete_directory` API on hierarchical namespace accounts can automatically delete the directory recursively with no need to explicitly list all the blobs it contains https://learn.microsoft.com/en-us/python/api/azure-storage-file-datalake/azure.storage.filedatalake.datalakedirectoryclient?view=azure-python#azure-storage-filedatalake-datalakedirectoryclient-delete-directory
1. Read the blob metadata to look for directory markers
    1. Directory markers on hierarchical namespace accounts have the metadata `Hdi_isfolder=true`. 
    2. This would be a simple and robust way to detect directory marker blobs that must be deleted in the correct order. 
    3. Unfortunately it will be a big performance cost to query the metadata of all these blobs unnecessarily. It seems like https://github.com/fsspec/adlfs/pull/383 added `expand_path=False` specifically to get this optimisation. 

### Testing
The currently existing test `test_rm_recursive` catches this bug if it is run against a hierarchical namespace storage account. I provisioned a real hierarchical namespace account to run this test against. 
I added an assertion to `test_rm_recursive` that the directory marker is the last delete. 
I also added a test that deletes more directories. To implement this I changed the `storage` fixture to have `function` scope to reset it after every test. This doesn't seem to impact the runtime of the tests so I assume this is ok.

### DIFF
--- a/adlfs/tests/conftest.py
+++ b/adlfs/tests/conftest.py
@@ -28,7 +28,7 @@ def host(request):
     return request.config.getoption("--host")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
 def storage(host):
     """
     Create blob using azurite.
@@ -56,6 +56,7 @@ def storage(host):
     container_client.upload_blob("root/e+f/file2.txt", data)
     yield bbs
 
+    bbs.delete_container("data")
 
 @pytest.fixture(scope="session")
 def event_loop():
--- a/adlfs/tests/conftest.py
+++ b/adlfs/tests/conftest.py
@@ -58,6 +58,7 @@ def storage(host):
 
     bbs.delete_container("data")
 
+
 @pytest.fixture(scope="session")
 def event_loop():
     policy = asyncio.get_event_loop_policy()
--- a/adlfs/tests/test_spec.py
+++ b/adlfs/tests/test_spec.py
@@ -751,6 +751,48 @@ def test_rm_recursive(mock_delete_blob, storage):
     ), "The directory deletion should be the last call"
 
 
+@mock.patch.object(
+    azure.storage.blob.aio.ContainerClient,
+    "delete_blob",
+    side_effect=azure.storage.blob.aio.ContainerClient.delete_blob,
+    autospec=True,
+)
+def test_rm_recursive2(mock_delete_blob, storage):
+    fs = AzureBlobFileSystem(
+        account_name=storage.account_name, connection_string=CONN_STR
+    )
+
+    assert "data/root" in fs.ls("/data")
+
+    # print(fs.ls("/data/", invalidate_cache=True))
+    # assert fs.ls("/data/*") == [
+    #     "data/root/a",
+    #     "data/root/a1",
+    #     "data/root/b",
+    #     "data/root/c",
+    #     "data/root/d",
+    #     "data/root/e+f",
+    #     "data/root/rfile.txt",
+    # ]
+    fs.rm("data/root", recursive=True)
+    assert "data/root" not in fs.ls("/data")
+
+    with pytest.raises(FileNotFoundError):
+        fs.ls("data/root")
+
+    last_deleted_paths = [call.args[1] for call in mock_delete_blob.mock_calls[-7:]]
+    print(last_deleted_paths)
+    assert last_deleted_paths == [
+        "root/e+f",
+        "root/d",
+        "root/c",
+        "root/b",
+        "root/a1",
+        "root/a",
+        "root",
+    ], "The directory deletion should be in reverse lexographical order"
+
+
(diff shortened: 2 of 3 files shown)


adlfs/tests/conftest.py
```diff
@@ -28,7 +28,7 @@ def host(request):
     return request.config.getoption("--host")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
```
Tom-Newton, line 31: This had no measurable impact on test runtime, therefore I'm hoping its ok to change this.

adlfs/tests/test_spec.py
```diff
@@ -401,10 +403,10 @@ def test_time_info(storage):
     )
 
     creation_time = fs.created("data/root/d/file_with_metadata.txt")
-    assert creation_time == storage.insert_time
+    assert_almost_equal(creation_time, storage.insert_time, datetime.timedelta(seconds=1))
 
     modified_time = fs.modified("data/root/d/file_with_metadata.txt")
-    assert modified_time == storage.insert_time
+    assert_almost_equal(modified_time, storage.insert_time, datetime.timedelta(seconds=1))
```
Tom-Newton, line 409: I saw this test flake, so I thought I would fix it.

TomAugspurger:
cc @daavoo do you have a chance to look at this?

adlfs/tests/test_spec.py
TomAugspurger: Looks like a small linting error. I  *think* this repo uses pre-commit, if you want to set that up locally.

Tom-Newton:
I think all the lint failures should be resolved now.

adlfs/tests/test_spec.py
```diff
@@ -738,6 +750,42 @@ def test_rm_recursive(storage):
...
+    assert "data/root" in fs.ls("/data")
+
+    fs.rm("data/root", recursive=True)
+    assert "data/root" not in fs.ls("/data")
+
+    with pytest.raises(FileNotFoundError):
+        fs.ls("data/root")
+
+    last_deleted_paths = [call.args[1] for call in mock_delete_blob.mock_calls[-7:]]
+    print(last_deleted_paths)
```
TomAugspurger, line 778:
```suggestion
```

TomAugspurger (approved):

TomAugspurger:
Looks good, thanks. Just removed a print statement in the tests. I'll merge once CI passes.

TomAugspurger:
Thanks @Tom-Newton!

efiop:
Thank you @Tom-Newton !

gdippolito:
Thanks a lot! @efiop would you folks have already a date in mind to cut a release containing this patch?

Issue #294, opened 2022-01-29
Set blob Content-Type in AzureBlobFileSystem.put()


jhamman:
Is it possible to set the `Content-Type` attribute (or other content settings) of blobs using adlfs? 

I would like to do something like this:

```python
fs = AzureBlobFileSystem(...)
fs.put('index.html', 'foo/bar', content_settings={'Content-Type': 'text/html'})
```

This does seem to be possible when using the azure storage library directly. For example:

```python
blob_client = ...
with open(index.html, "rb") as data:
    blob_client.upload_blob(data, content_settings=ContentSettings(content_type='text/html'))
```

I suspect this is possible today but the above example does not work so I'm looking for examples and/or documentation to support this sort of workflow.

cc @orianac

TomAugspurger:
That method take `**kwags` but silently ignores them. I think we could pass them through to `BlobClient.upload_blob` at https://github.com/fsspec/adlfs/blob/master/adlfs/spec.py#L1575-L1581.

IMO, it'd be best to just have users use the `azure.storage.blob` objects rather than trying to wrap them / infer what they mean from something like `content_settings={"Content-Type": ...}`. So I'd suggest just supporting

```
fs.put('index.html', 'foo/bar', content_settings=azure.storage.blob.ContentSettings(...))
```

by passing through kwargs unmodified.

isidentical:
Not sure if there are any complications to do this on azure blog storage, but just as passing-by note, s3fs does this automatically by guessing the mime type (and supports overriding, if you need a specific content type):
https://github.com/fsspec/s3fs/blob/736ee5ae7a16494e14fa12838cd963e1472afb9e/s3fs/core.py#L967-L970

TomAugspurger:
https://github.com/fsspec/filesystem_spec/issues/916 is an upstream issue for standardizing this. For now, I'm going to pass through `kwargs` so that we at least have the option to set it using `azure.storage.blob.ContentSettings`.

PR #456, opened 2024-01-29
Fix missing kwargs injection on put calls

rodrigoberriel:
While https://github.com/fsspec/adlfs/pull/392 was supposed to fix https://github.com/fsspec/adlfs/issues/294, it did not. When uploading a file using `fs.put`, `kwargs` is still not being forwarded, because these calls are passed to `_put_file` which does not rely on `_pipe`, unlike `write_*` calls.

This PR is very simple: also injects kwargs on `bc.upload_blob` for the `_put_file` calls.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1627,6 +1627,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
                             ),
                             max_concurrency=max_concurrency or self.max_concurrency,
                             **self._timeout_kwargs,
+                            **kwargs,
                         )
                 self.invalidate_cache()
             except ResourceExistsError:


rodrigoberriel:
@adam-roughton @TomAugspurger would you mind to consider merging this one for some of the next releases?

Without it, if someone wants to use put and set, for instance, tags and content-type for the blob, the user would have to issue two additional requests (`set_blob_tags` and `set_http_headers`) which adds latency and incurs in additional logging costs. I don't know if there's a workaround, but we currently have to rely on a internal fork only because of this simple change.

If the change is too small or you need anything else, please let me know.

karosc:
It also feels like there is some precedent with #392 getting merged. Why can we set tags using `fs.write_bytes`, but not with `fs.put`?

kyleknap:
@martindurant I'm ok with the change of proxying the kwargs to the `upload_blob()` method. I think my only hesitation is that it might be more robust to have a designated `upload_blob_kwargs` where a dictionary of the additional kwargs are provided e.g.
```python
fs.upload_file(
   'local-file',
   'blob-name',
   upload_blob_kwargs={
       "tags": {
           "tag1": "val1"
       }
   }
```
this would ensure that if we want to add any future keyword arguments that we expose won't collide with any Azure SDK keyword arguments, but it does make it more verbose. That being said, given `write_bytes()` already accepts SDK keyword arguments as top-level keyword arguments and the usage pattern is similar to `s3fs`, I do think it makes sense to stay consistent here. Thoughts?

> I think the point of maybe implementing the attr methods is also fair, although I don't know if you can simply change the metadata in azure (I assume you can).

Yeah tags can be set through separate API calls with the Azure SDK. I'm open to exposing it given `s3fs` offers this already, but I'd prefer we keep this PR to just proxying the keyword arguments, and we can revisit exposing a `get_tags` and `put_tags` method if still needed.

Otherwise, for this PR, it would be great if we could get a test added for both `put_file()` and `put()` to make sure the keyword arguments are passed to the SDK client call (e.g., setting tags) so that we definitively confirm that we've addressed: https://github.com/fsspec/adlfs/issues/294. @rodrigoberriel let us know if you have the bandwidth to update the PR. Otherwise, we can push some tests up and get it merged.

rodrigoberriel:
Sure. Today is kinda busy, but I can get it done by the weekend.

rodrigoberriel:
@kyleknap it does support it. I've updated the test.

kyleknap (approved):
@rodrigoberriel This looks good to me! I also confirmed that it was working correctly with tags when trying it out. A couple of updates we probably should add as well before merging:
* Let's rebase the PR off of main (or merging main into this branch works too). Mainly in order to get the tests to pass using the latest version of the Azure SDK there's a fix we made to the CI earlier this year that we'll want to pull in.
* Once updated with main, let's add a changelog entry for it [in this section](https://github.com/fsspec/adlfs/blob/main/CHANGELOG.md#unreleased). I think we can keep it to a sentence to call out that we proxy these keyword arguments now to the Azure Blob SDK client to support content settings, tags, etc.

Afterwords @martindurant I think we should be set to enable CI for the pull request to make sure CI passes and get the PR merged assuming you are good with the change too!

martindurant:
I enabled CI now, and can do again following changes.

kyleknap (approved):
Looks good to me! @martindurant we should be good to re-enable CI and get it merged.

Issue #459, opened 2024-02-08
Runtime warning from missing await


johnmacnamararseg:
Getting a Runtime Warning when calling `rm`. Seems like just need to add await in front on `delete_blob` [here](https://github.com/fsspec/adlfs/blob/576fb7a6a53a55375b4458c09e5bb571d945d410/adlfs/spec.py#L1223)

Error message: `spec.py:1223: RuntimeWarning: coroutine 'ContainerClient.delete_blob' was never awaited`

Appears this block was recently changed in this [PR](https://github.com/fsspec/adlfs/pull/454)

johnmacnamararseg:
Ill open a PR with a fix

johnmacnamararseg:
https://github.com/fsspec/adlfs/pull/460

PR #460, opened 2024-02-08
add missing await on delete_blob call per issue 459

johnmacnamararseg:
per https://github.com/fsspec/adlfs/issues/459

add missing await on `delete_blob`

### DIFF
--- a/adlfs/tests/test_spec.py
+++ b/adlfs/tests/test_spec.py
@@ -2,7 +2,6 @@ import datetime
 import tempfile
 from unittest import mock
 
-import azure.storage.blob
 import azure.storage.blob.aio
 import dask.dataframe as dd
 import numpy as np
@@ -754,14 +753,8 @@ def test_rm_recursive(mock_delete_blob, storage):
         mock.ANY, "root/c"
     ), "The directory deletion should be the last call"
 
-
-@mock.patch.object(
-    azure.storage.blob.aio.ContainerClient,
-    "delete_blob",
-    side_effect=azure.storage.blob.aio.ContainerClient.delete_blob,
-    autospec=True,
-)
-def test_rm_recursive2(mock_delete_blob, storage):
+@pytest.mark.filterwarnings("error")
+def test_rm_recursive2(storage):
     fs = AzureBlobFileSystem(
         account_name=storage.account_name, connection_string=CONN_STR
     )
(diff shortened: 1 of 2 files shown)


johnmacnamararseg:
tested this locally and appears to run fine. unsure if further testing is required

TomAugspurger:
Thanks, I think a unit test would be helpful to avoid future regressions. Can you add one?

johnmacnamararseg:
This issue might be a little more impactful than I originally thought. I initially thought we were just leaking the runtime warning but still cleaning up the directories. However, it looks like without await none of those `delete_blob` calls on the directory markers are executing - which makes sense since were attempting to execute these co-routines without `await`. Ran some local tests to confirm this.

I also didn't realize that this test suite could run out of the `azurite` container - pretty nifty. 

This change actually breaks [this unit test](`https://github.com/fsspec/adlfs/blob/576fb7a6a53a55375b4458c09e5bb571d945d410/adlfs/tests/test_spec.py#L764`)

It looks like this test and the one above it are mocking the `delete_blob` calls to assert that the various calls within `rm` run in a certain order and then setting the side affect to `delete_blob` so that the delete occurs. The reason this didn't fail the test originally and didn't produce the runtime warning was because the mocking + side effect turned the `delete_blob` call from a co-routine requiring the `await` into a conventional subroutine. 

All the files created in the `storage` harness are still deleted since the side effect `delete_blob` call exists outside of the async context manager in `_rm_files`. However, once I added `await` it doesn't seem like `mock_delete_blob` captures all of the `delete_blob` calls like we would expect. I don't yet have an explanation for this.

Going to see if I can modify these tests in a way where we can still evaluate the call ordering but the combination of Mocking + async is a bit out of my wheelhouse

adlfs/tests/test_spec.py
```diff
@@ -754,27 +753,51 @@ def test_rm_recursive(mock_delete_blob, storage):
...
     fs.rm("data/root", recursive=True)
+
     assert "data/root" not in fs.ls("/data")
 
     with pytest.raises(FileNotFoundError):
         fs.ls("data/root")
 
-    last_deleted_paths = [call.args[1] for call in mock_delete_blob.mock_calls[-7:]]
+
+async def test_rm_recursive_call_order(storage, mocker):
```
johnmacnamararseg, line 772: Since `fs.rm` validates that the files are actually removed, there wasn't an easy way to mock the `delete_blob` so that still ran the deletes asynchronously as a side effect (someone might have a clever way to do this). So instead I just explicitly supply the file paths under `data/root` , call `fs._rm_files`, and assert the call order similar to how the previous test was working

johnmacnamararseg:
@TomAugspurger changes pushed

johnmacnamararseg:
@TomAugspurger @efiop any chance we can get some eyes on this?

TomAugspurger:
Thanks @johnmacnamararseg!

nosterlu:
Thank you for fixing @johnmacnamararseg, I did not notice that folders were not deleted until you mentioned it. Looking forward to this release

PR #463, opened 2024-03-13
document that `credential` needs to be from azure.identity.aio

temporaer:
if it's a sync credential, then the weakref.finalize callback will cause an error, `close() -> None` cannot be awaited.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -147,7 +147,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         request session. See
         http://azure.microsoft.com/en-us/documentation/articles/storage-configure-connection-string/
         for the connection string format.
-    credential: azure.core.credentials.AsyncTokenCredential or SAS token
+    credential: azure.core.credentials_async.AsyncTokenCredential or SAS token
         The credentials with which to authenticate.  Optional if the account URL already has a SAS token.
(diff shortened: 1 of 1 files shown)


temporaer:
(There's a chance that this addresses the same issue as #330, though on the user side.)

adlfs/spec.py
```diff
@@ -147,9 +147,9 @@ class AzureBlobFileSystem(AsyncFileSystem):
         request session. See
         http://azure.microsoft.com/en-us/documentation/articles/storage-configure-connection-string/
         for the connection string format.
-    credential: TokenCredential or SAS token
+    credential: (async!) TokenCredential or SAS token
```
TomAugspurger, line 150:
```suggestion
    credential: azure.core.credentials.AsyncTokenCredential or SAS token
```

adlfs/spec.py
```diff
@@ -147,9 +147,9 @@ class AzureBlobFileSystem(AsyncFileSystem):
         request session. See
         http://azure.microsoft.com/en-us/documentation/articles/storage-configure-connection-string/
         for the connection string format.
-    credential: TokenCredential or SAS token
+    credential: azure.core.credentials.AsyncTokenCredential or SAS token
```
TomAugspurger, line 150:
```suggestion
    credential: azure.core.credentials_async.AsyncTokenCredential or SAS token
```

PR #468, opened 2024-04-15
Honor the anon parameter if set

### LINKED ISSUE #467: Parameter anon ignored if set to False
The change introduced by #437 has meant that the `anon` parameter is effectively ignored if set to `False`: `AzureBlobFileSystem(account_name, anon=False)`

Should be something like this:
```python
self.anon = anon if anon is not None else os.getenv("AZURE_STORAGE_ANON", "true").lower() not in [
    "false",
    "0",
    "f",
]
```
or perhaps more readable:
```python
if anon is not None:
   self.anon = anon
else:
   self.anon = os.getenv("AZURE_STORAGE_ANON", "true").lower() not in [
      "false",
      "0",
      "f",
  ]
```

adam-roughton:
If the anon parameter is set, it should be used over any environment variables set.

Fixes #467

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -277,11 +277,14 @@ class AzureBlobFileSystem(AsyncFileSystem):
         self.client_id = client_id or os.getenv("AZURE_STORAGE_CLIENT_ID")
         self.client_secret = client_secret or os.getenv("AZURE_STORAGE_CLIENT_SECRET")
         self.tenant_id = tenant_id or os.getenv("AZURE_STORAGE_TENANT_ID")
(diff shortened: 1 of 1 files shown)


microft (approved):

TomAugspurger:
Thanks. Could you add a regression test for this?On Apr 15, 2024, at 8:38 AM, Luís Miguel Braga ***@***.***> wrote:﻿
@microft approved this pull request.

TomAugspurger:
Going to merge this and cut a release, but if either of you could follow up adding a test that would be great.

PR #478, opened 2024-06-19
support signed urls via connection string alone

shcheklein:
In a lot of cases we have only a connection and no account name, account key. Those could be extracted from the connection string and it's convenient to have it done automatically by the fs itself.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -22,6 +22,7 @@ from azure.core.exceptions import (
     ResourceExistsError,
     ResourceNotFoundError,
 )
+from azure.core.utils import parse_connection_string
 from azure.storage.blob import (
     BlobBlock,
     BlobProperties,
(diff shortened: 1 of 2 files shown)


shcheklein:
@martindurant hey, sorry for bothering you here, do you know who is the right person to check this and do a release? thanks!

martindurant:
@hayesgb @TomAugspurger - who is maintaining this repo these days?

martindurant:
Is it possible to make a test for this change?

shcheklein:
> Is it possible to make a test for this change?

@martindurant yep, it has a test already

shcheklein:
@hayesgb @TomAugspurger @martindurant a friendly reminder folks :)

TomAugspurger:
It'll be a bit before I can take a look.
________________________________

adlfs/spec.py
```diff
@@ -1541,11 +1542,19 @@ async def _url(
         """
         container_name, blob, version_id = self.split_path(path)
 
+        if self.connection_string:
+            args_dict = parse_connection_string(self.connection_string)
```
martindurant, line 1546: OK, I see what's going on here, looks good to me. One assumes the connected client also has this information, but since there's an upstream function to parse it, I don't see why we shouldn't use it.

martindurant (approved):
I don't really do any maintenance on this project, so I hope someone else merges this.

adlfs/spec.py
shcheklein: yep, thanks @martindurant for approving ... I hope maintainers will be able to take a look soon :)

shcheklein:
just a friendly reminder folks, it would be great if we can merge this :)

TomAugspurger:
Thanks @shcheklein

shcheklein:
@TomAugspurger could you also trigger a new release please when you have a minute. Thanks!

TomAugspurger:
2024.7.0 is out on PyPI now: https://pypi.org/project/adlfs/

PR #480, opened 2024-06-21
Fix account host

dorbaker:
Added account_host parameter to AzureBlobFileSystem, which is needed for package to work in other Azure clouds.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -136,9 +136,10 @@ class AzureBlobFileSystem(AsyncFileSystem):
         If any of account key, sas token or client_id is specified, anonymous access
         will be used.
     account_host: str
-        The storage account host. This string is the entire url to the storage account
+        The storage account host. This string is the entire url to the for the storage
         after the https://, i.e. "https://{account_host}". This parameter is only
-        required for Azure clouds that do not end with "blob.core.windows.net".
+        required for Azure clouds where account urls do not end with "blob.core.windows.net". 
+        Note that the account_name parameter is still required.
     sas_token: str
         A shared access signature token to use to authenticate requests
         instead of the account key. If account key and sas token are both
@@ -239,8 +240,8 @@ class AzureBlobFileSystem(AsyncFileSystem):
     def __init__(
         self,
         account_name: str = None,
-        account_key: str = None,
         account_host: str = None,
+        account_key: str = None,
         connection_string: str = None,
         credential: str = None,
         sas_token: str = None,
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -135,11 +135,6 @@ class AzureBlobFileSystem(AsyncFileSystem):
         The storage account key. This is used for shared key authentication.
         If any of account key, sas token or client_id is specified, anonymous access
         will be used.
-    account_host: str
-        The storage account host. This string is the entire url to the for the storage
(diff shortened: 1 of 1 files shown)


adlfs/spec.py
```diff
@@ -287,6 +293,7 @@ def __init__(
             ]
         self.location_mode = location_mode
         self.credential = credential
+        if account_host: self.account_host = account_host
```
TomAugspurger, line 296: This will probably fail our linting on the formatting. You can run either `pre-commit run --all-files` or `black` locally to fix that.

adlfs/spec.py
```diff
@@ -235,6 +240,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
     def __init__(
         self,
         account_name: str = None,
+        account_host: str = None,
```
TomAugspurger, line 243: It's unfortunate, but we'll need to put this new parameter at the end. That way we don't break users using positional args `AzureBlobFileSystem(account_name, key, ...)`

adlfs/spec.py
```diff
@@ -135,6 +135,11 @@ class AzureBlobFileSystem(AsyncFileSystem):
         The storage account key. This is used for shared key authentication.
         If any of account key, sas token or client_id is specified, anonymous access
         will be used.
+    account_host: str
```
TomAugspurger, line 138: We'll have to be careful with the naming here.

Seems that Azure already uses "account host" to mean something slightly different. From https://learn.microsoft.com/en-us/azure/storage/blobs/storage-custom-domain-name?tabs=azure-portal#step-1-get-the-host-name-of-your-storage-endpoint

> The host name is the storage endpoint URL without the protocol identifier and the trailing slash.

I'd much prefer it if adlfs just took the `account_url` like the underlying `azure.storage.blob` library.

Then maybe in `__init__` we use

```
self._parsed = urllib.parse.urlparse(account_url)
```

which is something like

```python
In [4]: urllib.parse.urlparse("https://foo.blob.core.windows.net")
Out[4]: ParseResult(scheme='https', netloc='foo.blob.core.windows.net', path='', params='', query='', fragment='')
```

And then we store the pieces we need later on for `connect`, which I think is

- `scheme`
- `netloc.split(".", 1)[-1]`, e.g. `blob.core.windows.net`

Thoughts on that?

TomAugspurger (commented):
Thanks for the PR, left a few comments. LMK if you have time to look into them. Otherwise, no worries.

I suspect that this will be hard to test. We use azurite for emulating storage which uses its own account URL.

adlfs/spec.py
dorbaker: The account_host should be exactly the format you mentioned, as I attempted to describe in the parameter description. You can see that "https:/"/ is added in L481: "self.account_url: str = f"https://{self.account_host}".

I include the account_host parameter to mimic what's already in the code.

TomAugspurger: Oh sorry, read the docs you wrote complete backwards :)

adlfs/spec.py
dorbaker: I made it the last parameter so it doesn't break anything positional.

adlfs/spec.py
dorbaker: Done.

TomAugspurger:
Thanks!

Issue #462, opened 2024-03-04
Issue with parallel uploads to the same blob


M0dEx:
There seems to be an issue when 2 instances of this file system write to the same blob from 2 different processes in parallel, where one of the uploads fails with:

<details>
  <summary>Azure error</summary>

```
    File "/code/.venv/lib/python3.10/site-packages/our_package/connector/storage/blob.py", line 117, in _save
        with self._fs.open(
    File "/code/.venv/lib/python3.10/site-packages/fsspec/spec.py", line 1963, in __exit__
        self.close()
    File "/code/.venv/lib/python3.10/site-packages/adlfs/spec.py", line 1908, in close
        super().close()
    File "/code/.venv/lib/python3.10/site-packages/fsspec/spec.py", line 1930, in close
        self.flush(force=True)
    File "/code/.venv/lib/python3.10/site-packages/fsspec/spec.py", line 1801, in flush
        if self._upload_chunk(final=force) is not False:
    File "/code/.venv/lib/python3.10/site-packages/fsspec/asyn.py", line 118, in wrapper
        return sync(self.loop, func, *args, **kwargs)
    File "/code/.venv/lib/python3.10/site-packages/fsspec/asyn.py", line 103, in sync
        raise return_result
    File "/code/.venv/lib/python3.10/site-packages/fsspec/asyn.py", line 56, in _runner
        result[0] = await coro
    File "/code/.venv/lib/python3.10/site-packages/adlfs/spec.py", line 2068, in _async_upload_chunk
        await bc.commit_block_list(
    File "/code/.venv/lib/python3.10/site-packages/azure/core/tracing/decorator_async.py", line 77, in wrapper_use_tracer
        return await func(*args, **kwargs)
    File "/code/.venv/lib/python3.10/site-packages/azure/storage/blob/aio/_blob_client_async.py", line 1861, in commit_block_list
        process_storage_error(error)
    File "/code/.venv/lib/python3.10/site-packages/azure/storage/blob/_shared/response_handlers.py", line 184, in process_storage_error
        exec("raise error from None")   # pylint: disable=exec-used # nosec
    File "<string>", line 1, in <module>
    
azure.core.exceptions.HttpResponseError: The specified block list is invalid.
RequestId:<request_id>
Time:2024-02-13T12:15:05.1957595Z
ErrorCode:InvalidBlockList
[...]

PR #485, opened 2024-10-15
fix: #462 parallel uploads to the same blob storage

cmp0xff:
Closes https://github.com/fsspec/adlfs/issues/462.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -15,7 +15,8 @@ import weakref
 from collections import defaultdict
 from datetime import datetime, timedelta
 from glob import has_magic
-from typing import Optional, Tuple
+from typing import List, Optional, Tuple
+from uuid import uuid4
 
 from azure.core.exceptions import (
     HttpResponseError,
@@ -2138,8 +2139,7 @@ class AzureBlobFile(AbstractBufferedFile):
         """
         data = self.buffer.getvalue()
         length = len(data)
-        block_id = len(self._block_list)
-        block_id = f"{block_id:07d}"
+        block_id = self._get_block_id(self._block_list)
         if self.mode == "wb":
             try:
                 for chunk in self._get_chunks(data):
@@ -2152,8 +2152,7 @@ class AzureBlobFile(AbstractBufferedFile):
                             length=len(chunk),
                         )
                         self._block_list.append(block_id)
-                        block_id = len(self._block_list)
-                        block_id = f"{block_id:07d}"
+                        block_id = self._get_block_id(self._block_list)
 
                 if final:
                     block_list = [BlobBlock(_id) for _id in self._block_list]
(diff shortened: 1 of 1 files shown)


cmp0xff:
May I ask someone for a review? Maybe @TomAugspurger @martindurant since you reviewed the last PRs? Thank you in advance.

martindurant:
Is it possible to test this change, something that would fail without it?

cmp0xff:
Hi @martindurant , thank you for the reply. In the original issue #462, we discovered that
> There seems to be an issue when 2 instances of this file system write to the same blob from 2 different processes in parallel, where one of the uploads fails

I think it is not easy to write a pytest to show this issue explicitly.

martindurant:
I'm afraid I don't know anything about the function of these IDs... The problem was with assigning a constant ID to multiple uploads? I'm not sure I understand "who wins" in the case that two processes are writing at the same time.

martindurant:
Please add
```
from __future__ import annotations
```

Someone should also update the CI to a newer version of python, I would say at least 3.10.

cmp0xff:
Hi @martindurant ,

> I'm afraid I don't know anything about the function of these IDs... The problem was with assigning a constant ID to multiple uploads? I'm not sure I understand "who wins" in the case that two processes are writing at the same time.

The problem happens even when two _different_ blocks are being written to the blob at the same time, because the original implementation only take into account the _number_ of blocks. This is the case we want to solve here.

> Please add
> 
> ```py
> from __future__ import annotations
> ```
> 
> Someone should also update the CI to a newer version of python, I would say at least 3.10.

https://github.com/fsspec/adlfs/commit/d83487ada03b5fd437cd64fbb1f67fe316522f6a should have fixed the pipeline for py38. Please run the pipeline again, thanks.

martindurant:
Rather than hashing the data (which might be expensive), would any random value do?

adlfs/spec.py
```diff
@@ -2168,7 +2167,7 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
                 # which is throws an InvalidHeader error from Azure, so instead
                 # of staging a block, we directly upload the empty blob
                 # This isn't actually tested, since Azureite behaves differently.
-                if block_id == "0000000" and length == 0 and final:
+                if block_id == self._get_block_id([]) and length == 0 and final:
```
cmp0xff, line 2170: Hi, I've changed the implementation.

Now, if `block_list` is empty, the `block_id` is `"0" * 32`; otherwise, `uuid4().hex`. Both are strings of [length 32](https://docs.python.org/3/library/uuid.html#uuid.UUID.hex).

Also, empty block lists always map to the same result `"0" * 32`, which should keep this line work.

martindurant: That sounds like the right solution to me.

PR #488, opened 2024-11-14
update versions

martindurant:


cmp0xff:
https://github.com/martindurant/adlfs/blob/569f9196956bf1cfef66f49d13650af19a2f1dfa/adlfs/spec.py#L1569

It seems we can write
```py
            expiry=datetime.datetime.now(datetime.timezone.utc) + timedelta(seconds=expires),
```
The proposal comes from the fact that most of the warning
```
DeprecationWarning: Testing an element's truth value will raise an exception in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
```
is related to `datetime.utcnow()`, from my Google search.

martindurant:
I already fixed one [here](https://github.com/fsspec/adlfs/pull/488/files#diff-21aceffb2157c06c8d26699af7cacb074cd7fd5684d880f881e38f93e1b7f76fR42). 

But there are still some warnings coming from dep packages, which I think we should ignore for the "earliest" run.

.github/workflows/tests.yml
```diff
@@ -22,8 +22,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04]
-        pyv: ["3.8"]
+        os: [ubuntu-22.04]
+        pyv: ["3.12"]
```
cmp0xff, line 26: I am not an expert in devops, but maybe we want to run tests on older Python versions as well? py39 [is still being supported](https://devguide.python.org/versions/).
```suggestion
        pyv: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```

cmp0xff (commented):
And as @martindurant has written in https://github.com/fsspec/adlfs/pull/488#issuecomment-2479026732, maybe for py312 we need to xfail some tests for the `earliest` requirements.

.github/workflows/tests.yml
martindurant: That would make 5x3=15 CI runs. Probably we should test all those python versions with "earliest", but only 3.12 for "latest" and "dev".

.github/workflows/tests.yml
```diff
@@ -22,8 +22,8 @@ jobs:
...
         reqs: ["earliest", "latest", "dev"]
```
cmp0xff, line 27: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-adding-configurations
```suggestion
        os: [ubuntu-22.04]
        pyv: ["3.9", "3.10", "3.11", "3.12", "3.13"]
        reqs: [earliest]
        include:
          - pyv: "3.12"
            reqs: latest
          - pyv: "3.12"
            reqs: dev
```

cmp0xff: Update: in the current `earliest` requirements, many dependency versions officially support up to `py310`. I guess we should probably test all versions with `latest`, and test `py310` with `earliest` and `dev`.

requirements/earliest.txt
```diff
@@ -1,5 +1,5 @@
 -r base.txt
-fsspec==2023.9.0
-azure-core==1.23.1
-azure-datalake-store==0.0.46
-azure-storage-blob==12.12.0
+fsspec==2023.12.0
+azure-core==1.28.0
+azure-datalake-store==0.0.53
+azure-storage-blob==12.17.0
```
cmp0xff, line 5: - [fsspec 2023.12.0](https://pypi.org/project/fsspec/2023.12.0/): officially supports up to `py311`
- [azure-storage-blob 12.17.0](https://pypi.org/project/azure-storage-blob/12.17.0/): officially supports up to `py310`

adlfs/spec.py
```diff
@@ -1233,9 +1233,12 @@ async def _rm_files(
...
             # asynchronously for performance reasons.
-            file_exs = await asyncio.gather(
-                *([cc.delete_blob(file) for file in files]), return_exceptions=True
-            )
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+
+                file_exs = await asyncio.gather(
+                    *([cc.delete_blob(file) for file in files]), return_exceptions=True
+                )
```
cmp0xff, line 1241: If there is another `warnings.simplefilter("error")` deeply hidden inside `cc.delete_blob`, our `simplefilter("ignore")` will possibly be overridden. This could possibly be the reason why it did not fix the pipeline.

requirements/earliest.txt
martindurant: Usually, they will support more recent too, but at the time of release they were not explicitly tested with a later python (they come quite frequently these days).

PR #493, opened 2025-02-14
Also strip 'wasb' and 'wasbs' protocol schemes

christophediprima:
Add support to wasb scheme.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -374,7 +374,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
         STORE_SUFFIX = ".dfs.core.windows.net"
         logger.debug(f"_strip_protocol for {path}")
-        if not path.startswith(("abfs://", "az://", "abfss://")):
+        if not path.startswith(("abfs://", "az://", "abfss://", "wasb://", "wasbs://")):
             path = path.lstrip("/")
             path = "abfs://" + path
         ops = infer_storage_options(path)


christophediprima:
Can this be merged? This has been tested on both Azurite and Azure Blob Storage.

heman026:
Any update on this? When this will be merged

martindurant:
@kyleknap ?

kyleknap:
@christophediprima @heman026 Could you elaborate on why this is needed? And what is this change blocking? I saw there's quite a few cross-linked GitHub issues on the Iceberg repository, but I'm not sure I'm able to fully untangle the line of reasoning for needing this change.

My main hesitations right now are:
* Adlfs already accepts several schemes, and I'd like to avoid expanding the number of schemes accepted for simplicity and maintainability reasons.
* Support for `wasb` and `wasbs` protocols seems to refer to the WASB driver: https://hadoop.apache.org/docs/stable/hadoop-azure/wasb.html which is on the path for deprecation and removal.

That being said if the motivation is strong for supporting this, I'm open to pulling it in.

kevinjqliu:
hey @kyleknap chiming in from the (py)iceberg side. We use both pyarrow and fsspec for filesystem operations. It would be great for the 2 to have feature parity regarding the supported protocols. 

[Pyarrow AzureFileSystem](https://arrow.apache.org/docs/python/generated/pyarrow.fs.AzureFileSystem.html#pyarrow.fs.AzureFileSystem) added support for these parameters
```
blob_storage_authority str, default None
hostname[:port] of the Blob Service. Defaults to .blob.core.windows.net. Useful for connecting to a local emulator, like Azurite.

blob_storage_scheme str, default None
Either http or https. Defaults to https. Useful for connecting to a local emulator, like Azurite.

dfs_storage_authority str, default None 
hostname[:port] of the Data Lake Gen 2 Service. Defaults to .dfs.core.windows.net. Useful for connecting to a local emulator, like Azurite. 

dfs_storage_scheme str, default None 
Either http or https. Defaults to https. Useful for connecting to a local emulator, like Azurite.
```
This allows us to test using the different protocol against azurite.

kevinjqliu:
regarding your concerns above, 

> Adlfs already accepts several schemes, and I'd like to avoid expanding the number of schemes accepted for simplicity and maintainability reasons.

+1, i would recommend the approach pyarrow took. keep the same defaults but make it configurable. 

> Support for wasb and wasbs protocols seems to refer to the WASB driver: https://hadoop.apache.org/docs/stable/hadoop-azure/wasb.html which is on the path for deprecation and removal.

seems weird to add a feature that is deprecated for removal, but for now id like to be feature parity with pyarrow fs

Let me know what you think!

kevinjqliu:
Caught up with @kyleknap offline. 

There are a couple of different protocol schemes for azure storage. Amongst those are abfs[s], az, and wasb[s]. These are written depending on the specific libraries used. For wasb[s],  any client that uses https://hadoop.apache.org/docs/stable/hadoop-azure/wasb.html will write wasb[s] as the uri. Snowflake is also writing this scheme (see https://github.com/apache/iceberg-python/issues/1606 and https://github.com/apache/iceberg/issues/10127)

As a filesystem implementation for azure storage, adlfs should also support wasb[s]. The support here only means to allow parsing this scheme and its related uri. There is enough information in the wasb[s] uri to be used by the underlying adlfs to interact with storage. 

We should add wasb[s] support to pyarrow as well.

PR #500, opened 2025-06-19
Fixed flake8 error in ci

anjaliratnam-msft:
Resolves the flake8 error where the nonlocal sent_total variable is not assigned a value by adding the ignore code noqa: f824.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -78,16 +78,12 @@ def make_callback(key, callback):
     if callback is None:
         return None
 
-    sent_total = False
-
     def wrapper(response):
-        nonlocal sent_total # NOQA: F824
-
         current = response.context.get(key)
         total = response.context["data_stream_total"]
         if current is None:
             return
-        if not sent_total and total is not None:
+        if total is not None:
             callback.set_size(total)
         callback.absolute_update(current)
 


anjaliratnam-msft:
To fix the flake8 error, I removed the sent_total variable which is never updated. This is likely a bug and a follow up pr will be sent to address the issue that the size is getting set every time it is called.

anjaliratnam-msft:
Here is the workflow run where all the tests pass: https://github.com/anjaliratnam-msft/adlfs/actions/runs/15888307068

kyleknap (approved):
Looks good :ship:

@martindurant would you be able to enable CI for the pull request and give a review of it in order to get CI passing again for `main`?

For context on the approach, it looks like `flake8` started failing in CI because `flake8` released an update where it detects unused `nonlocal`/`global` declarations: https://github.com/PyCQA/flake8/pull/1976 and flagged this instance where the code never sets the `sent_total` to `True`. As @anjaliratnam-msft noted this is likely the unwanted behavior; it means `adlfs` callbacks are currently invoked twice for every update: once for setting the total and once for setting the update. Instead, the desired behavior it seems is to update the expected total only once.

We opted to just remove the `sent_total` in order to unblock CI and avoid introducing any new behaviors for now. As @anjaliratnam-msft mentioned, we'd like to circle back on this once CI is in a stable state assuming you agree that the total size should only be set once for fsspec callbacks?  There's also some concern that `adlfs` is using a legacy interface for tracking progress and can run into concurrency issues in reporting similar to what the Azure CLI experienced: https://github.com/Azure/azure-cli/issues/21781. There is a new `progress_callback` parameter to Azure SDK client that is now recommended and addresses the concurrency issues. We want to confirm if the issue listed in the Azure CLI also impacts `adlfs` and address it as part of circling back on progress reporting fixes. 

Thoughts?

TomAugspurger:
I triggered the workflow run.

LMK if you all are interested in maintaining adlfs. I don't have time for it these days.

PR #501, opened 2025-06-23
adlfs user agent

anjaliratnam-msft:
Added adlfs to the user agent to differentiate between adlfs and azure python sdk usage. 

Request headers example:
    'x-ms-version': 'REDACTED'
    'Accept': 'application/xml'
    'User-Agent': 'adlfs/2024.12.0 azsdk-python-storage-blob/12.25.1 Python/3.13.5 (Windows-11-10.0.26100-SP0)'
    'x-ms-date': 'REDACTED'
    'x-ms-client-request-id': '0fbad80d-5059-11f0-9c34-000d3a6d20b2'
    'Authorization': 'REDACTED'


adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2048,15 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
         with fs.open("data/afile", "xb") as f:
             pass
     assert fs.cat_file("data/afile") == b"data"
+
+
+def test_user_agent(storage: azure.storage.blob.BlobServiceClient, mocker):
```
kyleknap, line 2053: It would be great if we could expand + parameterize this test case to assert user agent being set for:
1. Both the file system and file classes
2. Both client creation approaches i.e., using `from_connection_string` and using the initializer

I think this is important given the amount of branching logic when it comes to client creation so that there is coverage for the various ways clients are created in `adlfs`

kyleknap (commented):
Looks good. I just had a couple of suggestions.

adlfs/spec.py
```diff
@@ -118,6 +124,41 @@ def _coalesce_version_id(*args) -> Optional[str]:
...
+            account_url=account_url,
+            credential=None,
+            _location_mode=location_mode,
+            user_agent=_get_user_agent(),
+        )
+    else:
+        return AIOBlobServiceClient(
+            account_url=account_url,
+            user_agent=_get_user_agent(),
+        )
```
kyleknap, line 150: For this factory, it would be interesting to see if we can consolidate the actual client instantiation to one instance. To do this we should be able to leverage keyword arguments, for example:
```python
service_client_kwargs = {
    "account_url": account_url,
    "user_agent": _get_user_agent(),
}
if credential is not None:
    service_client_kwargs["credentials"] = credentials
if location_mode is not None:
    service_client_kwargs["_location_mode"] = location_mode
return AIOBlobServiceClient(**service_client_kwargs)
```
The main reason that I'm suggesting we do this is it reduces the verbosity and makes it easier to follow/add to the logic when there is only one path to instantiating the client.

adlfs/spec.py
```diff
@@ -118,6 +124,41 @@ def _coalesce_version_id(*args) -> Optional[str]:
         return version_ids.pop()
 
 
+def _create_aio_blob_service_client(
+    account_url: str,
+    location_mode: Optional[str] = None,
+    credential: Optional[str] = None,
```
kyleknap, line 130: Hmm technically this is not the correct type; it can be more than a string. But it looks like the initializer for the filesystem use this type as well. I think it is fine to keep it as is but it may be worth doing a pass at a later time to add correct typing to the project and integrate `mypy` into the CI. Mainly the value add is add another level of safety when adding to the project and it allows consumers to use correct types when consuming adlfs.

adlfs/spec.py
```diff
@@ -73,6 +73,12 @@
 _SOCKET_TIMEOUT_DEFAULT = object()
 
 
+def _get_user_agent():
```
kyleknap, line 76: In order to break the circular loop here, I think it would be worth moving this logic: https://github.com/anjaliratnam-msft/adlfs/blob/f64dff2b2ce62dd2d194e0be02804104b4505d7d/adlfs/__init__.py#L6-L11 to the `utils.py` module and then have the `__init__.py` import `__version__` and `version_tuple` and this module can import `__version__`. Then we can go back to having a module level `_USER_AGENT` constant.

Generally, a `utils.py` module will have all of the shared code and not have any intra project imports which is usually a safe approach to avoiding import errors.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,100 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+        account_name=storage.account_name,
+    )
+    AzureBlobFile(fs, "data/root/a/file.txt", mode="rb")
+
+    mock_client.assert_called_once()
+    assert "user_agent" in mock_client.call_args.kwargs
+    assert mock_client.call_args.kwargs["user_agent"] == f"adlfs/{__version__}"
+
+
+@patch("adlfs.spec.AIOBlobServiceClient")
```
kyleknap, line 2116: Could you elaborate on why this patch outside of the `mocker` is needed? I'm not sure how well the patch will behave when we patch the class here and then patch the `from_connection_string` of the already patched class later on. I'd assume we'd only need one of the patches for the purpose introspection and mocking.

kyleknap (commented):
This is looking good. I mainly focused on the implementation. I still want to take a deeper pass on the tests to understand how the scaffolding and fixtures are set up with adlfs as part of reviewing the PR's tests.

adlfs/tests/test_spec.py
kyleknap: Yeah that sounds like a patch that is not properly being cleaned up. I'll take a look to see if I can find anything

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,100 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+    )
+    mock_client.return_value.get_container_client = mocker.MagicMock()
+
+    f = AzureBlobFile(fs, "data/root/a/file.txt", mode="rb")
+    f.container_name = "data"
+    f.connect_client()
+
+    mock_client.assert_called_once()
+    assert "user_agent" in mock_client.call_args.kwargs
+    assert mock_client.call_args.kwargs["user_agent"] == f"adlfs/{__version__}"
```
kyleknap, line 2071: It would probably make sense to consolidate these assertions to a helper function (e.g., `assert_sets_adlfs_user_agent`) and we can just provide it the mock client method whether it be the mock initializer or the `from_connection_string` mock.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,100 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+    storage: azure.storage.blob.BlobServiceClient, mocker
+):
+    from adlfs import __version__
+
+    fs = AzureBlobFileSystem(
+        account_name=storage.account_name, connection_string=CONN_STR
+    )
+    mock_client = mocker.patch.object(
+        AIOBlobServiceClient, "from_connection_string", return_value=mocker.MagicMock()
+    )
```
kyleknap, line 2062: Playing around with these mocks, I think it would be worth just having two new patch fixtures: one for the `__init__` and  one for the `from_connection_string()` method e.g.,:
```python
    @pytest.fixture()
    def mock_from_connection_string(mocker):
        return mocker.patch.object(
            AIOBlobServiceClient, "from_connection_string", autospec=True,
            side_effect=AIOBlobServiceClient.from_connection_string
        )


    @pytest.fixture()
    def mock_service_client_init(mocker):
        return mocker.patch.object(
            AIOBlobServiceClient, "__init__", autospec=True,
            side_effect=AIOBlobServiceClient.__init__
        )
```

This is a pattern used in another test case in `test_spec.py` where it effective will allow us to spy the arguments passed in but delegated to the actual method. Right now, in patching the entire class or building up mocks seems like a lot of scaffolding would be required to be able to get a test passing and I'm not sure if the scaffolding makes sense when we are already leveraging Azurite for the fake. With this approach, we should just be able to get what we want, which is capturing what was passed to the various class instantiation methods.

adlfs/spec.py
```diff
@@ -73,6 +73,12 @@
 _SOCKET_TIMEOUT_DEFAULT = object()
```
kyleknap, line 73: We should also make sure to add a changelog entry for this update here: https://github.com/fsspec/adlfs/blob/main/CHANGELOG.md

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,100 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+    from adlfs import __version__
+
+    mock_client_instance = mocker.MagicMock()
+    mock_client_instance.close = mocker.AsyncMock()
+    mock_client = mocker.patch(
+        "adlfs.spec.AIOBlobServiceClient.from_connection_string",
+        return_value=mock_client_instance,
+    )
+
+    AzureBlobFileSystem(account_name=storage.account_name, connection_string=CONN_STR)
```
kyleknap, line 2129: Whenever creating an `AzureBlobFileSystem`, we should definitely be setting ` skip_instance_cache=True` so that we make sure we always construct a new file system for these cases around instantiation

adlfs/tests/test_user_agent.py
```diff
@@ -0,0 +1,85 @@
...
+from azure.storage.blob.aio import BlobServiceClient as AIOBlobServiceClient
+
+from adlfs import AzureBlobFile, AzureBlobFileSystem
+from adlfs.tests.constants import CONN_STR, KEY
+from adlfs.utils import __version__ as __version__
+
+
+def assert_sets_adlfs_user_agent(mock_client_create_method):
+    mock_client_create_method.assert_called()
+    assert "user_agent" in mock_client_create_method.call_args.kwargs
```
kyleknap, line 12: The `call_args` property returns only the last call to the mock. So, we may still have the problem where we are only asserting the call on the client created in the file system instead of the file object if the file object just reuses the client from the file system. Instead, we should probably:

1. Leverage [`call_args_list`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args_list) to ensure all calls made include the user agent. 
2. Add an `expected_call_count` with a default of `1` to this helper assertion so that for the file-related tests we can assert that the client was created twice; once in the filesystem and once in the file-like object as there is no way to avoid avoid creating the client in the file system object.

kyleknap (commented):
Looks good. Just had one more comment on the test cases.

kyleknap (approved):

adlfs/spec.py
```diff
@@ -118,6 +121,31 @@ def _coalesce_version_id(*args) -> Optional[str]:
...
+        service_client_kwargs["credential"] = credential
+    if location_mode is not None:
+        service_client_kwargs["_location_mode"] = location_mode
+    return AIOBlobServiceClient(**service_client_kwargs)
+
+
+def _create_aio_blob_service_client_from_connection_string(
+    connection_string: str,
+) -> AIOBlobServiceClient:
+    return AIOBlobServiceClient.from_connection_string(
```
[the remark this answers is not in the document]
kyleknap: Gotcha. Yeah I'd imagine with botocore/aiobotocore creating a client would need to have async involved; there's quite a bit of I/O when a botocore client gets created (e.g., reading model files from disk to dynamically create the client class, reading configuration from disk, etc.). The async azure SDK client does not really have any I/O operations as part of client creation and so it does not use coroutines to instantiate it. So, I think the current approach to client creation (before and after) this PR should be fine.

Issue #497, opened 2025-05-27
Not working implementation of AzureBlobFileSystem.rm_file


frantisek-tichy-sf:
when trying to run this code, `NotImplementedError` arises when trying to call the `rm_file` method:
```py
import adlfs

STORAGE_ACCOUNT = "yourstorageaccount"
FILE_PATH = "path/to/your/file"

b = adlfs.AzureBlobFileSystem(STORAGE_ACCOUNT, anon=False)
b.rm_file(FILE_PATH)
print("done")
```

even though that in the `rm` method on the object works fine and when exploring the call stack, the `rm_file` figures in the stack.
Try for yourself: replace `rm_file` for `rm` on the second to last row

# Expected behavior #
file gets removed from the blob storage, the same way as it works with `rm` method.

PR #503, opened 2025-07-09
Set rm_files to be a synchronous method

anjaliratnam-msft:
This addresses the [github issue](https://github.com/fsspec/adlfs/issues/497) where rm_files is not implemented. This was a simple fix where [sync_wrapper(_rm_files)](https://github.com/fsspec/adlfs/blob/main/adlfs/spec.py#L1251) just needed to be set to rm_files. Tests were also added to make sure it works as expected.

### DIFF
--- a/adlfs/tests/test_spec.py
+++ b/adlfs/tests/test_spec.py
@@ -2237,15 +2237,10 @@ def test_rm_file_versioned_blob(storage, mocker):
         connection_string=CONN_STR,
         version_aware=True,
     )
-
     mock_delete_blob = mocker.patch.object(
         ContainerClient, "delete_blob", return_value=None
     )
     path = f"data/test_file.txt?versionid={DEFAULT_VERSION_ID}"
-    with fs.open(path, "wb") as f:
-        f.write(b"test content")
-
-    assert fs.exists(path)
     fs.rm_file(path)
     mock_delete_blob.assert_called_once_with(
         "test_file.txt", version_id=DEFAULT_VERSION_ID


adlfs/spec.py
```diff
@@ -1248,7 +1248,7 @@ async def _rm_files(
         for file in file_paths:
             self.invalidate_cache(self._parent(file))
 
-    sync_wrapper(_rm_files)
+    rm_files = sync_wrapper(_rm_files)
```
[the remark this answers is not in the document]
anjaliratnam-msft: This was the previous implementation of [rm_file](https://github.com/fsspec/adlfs/blob/6141bdc7391f2825e2ce8b7e1ddc722eb9c40e66/adlfs/spec.py#L1207) I found, but it looks like it was only ever used by rm and was not callable.

adlfs/spec.py
```diff
@@ -1278,6 +1278,31 @@ async def _rm_files(
...
+        ----------
+        path: str
+            File to delete.
+        """
+        container_name, p, _ = self.split_path(path, delimiter=delimiter)
+        try:
+            if p != "":
+                await self._rm_files(container_name, [p.rstrip(delimiter)])
+            else:
+                await self._rmdir(container_name)
```
kyleknap, line 1296: So looking more at [`rm_file`](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.rm_file) it seems like it's sole purpose is to just delete a single file and does not support deleting a directory provided. This seems to be the intention based on other implementation's versions of `rm_file` as well:

* [`s3fs`](https://github.com/fsspec/s3fs/blob/d46c149d89590b4b9109213f9f92cea01504eb0b/s3fs/core.py#L2052-L2059) - Only deletes a single object which would be a blob in adlfs
* [`local`](https://github.com/fsspec/filesystem_spec/blob/cf798ec684e0583906dda1ae6ff40de620c01eca/fsspec/implementations/local.py#L185) - Uses `os.remove()` which only handles removing files and not directories.

I think it would make sense to stick with this contract to be consistent, especially if `rm_file` was not actually never exposed publicly.

adlfs/spec.py
```diff
@@ -1278,6 +1278,31 @@ async def _rm_files(
 
     sync_wrapper(_rm_files)
 
+    async def _rm_file(
+        self, path: typing.Union[str, typing.List[str]], delimiter: str = "/", **kwargs
```
kyleknap, line 1282: And on a similar note, I'm not sure if we should be exposing a `delimiter` for this method since there is not really any recursive nature to this public contract and instead always just use `/` if we need any splitting logic. Also not supporting a `delimiter` seems consistent with the other implementations I linked.

kyleknap (commented):
It's looking better. Just had a couple of follow up suggestions on the direction we take this.

kyleknap (commented):
I think we are close. Just a few more comments.

kyleknap (approved):
Looks good! :ship:

@martindurant this should be ready for you to take a look

Issue #505, opened 2025-07-19
Azurite (currently) doesn't support latest API


jamesmyatt:
Current test failures look like they're caused by Azurite rather than adlfs: https://github.com/Azure/Azurite/issues/2562

Note sure what the preferred approach is. Maybe constraining the azure SDK libraries in CI. I think we can assume that the latest version works if the previous one does.

PR #506, opened 2025-07-21
Added skipApiVersionCheck for Azurite

anjaliratnam-msft:
Added the skipApiVersionCheck flag to Azurite to get the tests to pass since Azurite does not support the [latest API version](https://github.com/fsspec/adlfs/issues/505).

### DIFF
--- a/adlfs/tests/conftest.py
+++ b/adlfs/tests/conftest.py
@@ -65,7 +65,7 @@ def spawn_azurite():
     client = docker.from_env()
     azurite = client.containers.run(
         "mcr.microsoft.com/azure-storage/azurite",
-        "azurite-blob --loose --blobHost 0.0.0.0",
+        "azurite-blob --loose --blobHost 0.0.0.0 --skipApiVersionCheck",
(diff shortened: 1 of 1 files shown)


kyleknap (approved):
Looks good to me! :ship: 

I think just skipping the API version check in the long term makes sense as even if Azurite does eventually support this API version, there's no guarantee the Azurite will not lag in support for future API versions and break CI again.

It is also better than some of the other alternatives considered:
1. Pinning the version of the blob SDK. This is not great because we would not be testing adlfs on the latest version of the SDK which will likely be used by most adlfs customers. Also pinning it in the adlfs package could cause dependency resolution issues if other packages depend on the SDK.
2. Pinning the API version used by the SDK client. This is not ideal either because it could mean we lose out in accessing Storage API features that are gated by the API version.
3. Remove Azurite as a testing dependency. This is something we can consider in the future if the dependency continues to be an issue, but would require a heavy amount of refactoring.

martindurant:
+1 to unblocking work!

PR #507, opened 2025-07-24
Introduce connect client tests

kyleknap:
This builds off of the tests in this PR: https://github.com/fsspec/adlfs/pull/501 that asserted adlfs user agents are set for SDK clients. Even though the user agent tests added some coverage to how adlfs connected to SDK coverage, the adlfs test suite lacked coverage for how adlfs configurations were proxied to SDK clients (e.g., which account url to use and what credentials to set). By having these tests, it will make it easier to correctly do refactorings related to client creation (e.g., this PR: https://github.com/fsspec/adlfs/pull/504) and easier to add support new SDK client configurations in the future.

In addition to add these tests, a bug was discovered where for SAS token authentication on the file-like object, it would not respect location mode, while the file system did. The logic in the file-like object was updated to be consistent with files system logic since SAS tokens can be reused for secondary endpoints.

### DIFF
--- a/adlfs/tests/test_connect_client.py
+++ b/adlfs/tests/test_connect_client.py
@@ -33,30 +33,22 @@ def get_expected_client_init_call(
     account_url,
     credential=None,
     location_mode="primary",
-    user_agent=f"adlfs/{__version__}",
 ):
     call_kwargs = {
         "account_url": account_url,
+        "user_agent": f"adlfs/{__version__}",
     }
     if credential is not None:
         call_kwargs["credential"] = credential
     if location_mode is not None:
         call_kwargs["_location_mode"] = location_mode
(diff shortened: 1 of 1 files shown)


adlfs/tests/test_connect_client.py
```diff
@@ -0,0 +1,205 @@
...
+    user_agent=f"adlfs/{__version__}",
+):
+    call_kwargs = {
+        "account_url": account_url,
+    }
+    if credential is not None:
+        call_kwargs["credential"] = credential
+    if location_mode is not None:
+        call_kwargs["_location_mode"] = location_mode
+    if user_agent is not None:
```
anjaliratnam-msft, line 45: If the user agent is never none, does there need to be an if statement?

PR #508, opened 2025-07-25
Faster file-like writes

anjaliratnam-msft:
Utilized the max_concurrency argument to allow concurrent writes in the _async_upload_chunk function.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -2153,17 +2153,18 @@ class AzureBlobFile(AbstractBufferedFile):
         length = len(data)
         while start < length:
             end = min(start + self.blocksize, length)
-            yield data[start:end]
+            yield start, end
             start = end
 
-    async def _upload(self, chunk, block_id, semaphore):
+    async def _stage_block(self, data, start, end, block_id, semaphore):
         async with semaphore:
             async with self.container_client.get_blob_client(blob=self.blob) as bc:
                 await bc.stage_block(
                     block_id=block_id,
-                    data=chunk,
-                    length=len(chunk),
+                    data=data[start:end],
+                    length=end - start,
                 )
+                return block_id
 
     async def _async_upload_chunk(self, final: bool = False, **kwargs):
         """
@@ -2176,30 +2177,26 @@ class AzureBlobFile(AbstractBufferedFile):
             self.autocommit is True.
 
         """
-        data = self.buffer.getvalue()
-        length = len(data)
         block_id = self._get_block_id(self._block_list)
+        if len(self._block_list) == 0:
+            self._block_list.append(block_id)
         commit_kw = {}
         if self.mode == "xb":
             commit_kw["headers"] = {"If-None-Match": "*"}
         if self.mode in {"wb", "xb"}:
+            data = memoryview(self.buffer.getvalue())
+            length = len(data)
             try:
-                max_concurrency = self.fs.max_concurrency
+                max_concurrency = self.fs.max_concurrency or 1
                 semaphore = asyncio.Semaphore(max_concurrency)
                 tasks = []
-                block_ids = self._block_list or []
-                start_idx = len(block_ids)
-                chunks = list(self._get_chunks(data))
-                for _ in range(len(chunks)):
-                    block_ids.append(block_id)
-                    block_id = self._get_block_id(block_ids)
-
-                if chunks:
-                    self._block_list = block_ids
-                for chunk, block_id in zip(chunks, block_ids[start_idx:]):
-                    tasks.append(self._upload(chunk, block_id, semaphore))
-
-                await asyncio.gather(*tasks)
+                for start, end in self._get_chunks(data):
+                    tasks.append(
+                        self._stage_block(data, start, end, block_id, semaphore)
+                    )
+                    block_id = self._get_block_id(self._block_list)
+                ids = await asyncio.gather(*tasks, return_exceptions=True)
+                self._block_list.extend(ids[1:])
 
                 if final:
                     block_list = [BlobBlock(_id) for _id in self._block_list]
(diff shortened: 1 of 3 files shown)


adlfs/spec.py
```diff
@@ -2172,18 +2174,31 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
...
+                block_ids = []
+                chunks = list(self._get_chunks(data))
+                for _ in range(len(chunks)):
+                    block_ids.append(block_id)
+                    block_id = self._get_block_id(block_ids)
+                if chunks:
+                    self._block_list = block_ids
+                for chunk, block_id in zip(chunks, block_ids):
+
+                    async def _upload_chunk(chunk=chunk, block_id=block_id):
```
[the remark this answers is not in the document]
martindurant: +1, thanks for letting me know.

> the sizes used in s3fs of 50 MiB

The old default was 5MB, and you'll still see this in some places in fsspec. A good choice for the previous generation of small bandwidth, poor for today, especially if the compute is colocated with the storage.

adlfs/spec.py
```diff
@@ -2167,25 +2178,29 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
             self.autocommit is True.
 
         """
-        data = self.buffer.getvalue()
+        if self._sdk_supports_memoryview_for_writes() and self.mode in {"wb", "xb"}:
+            data = memoryview(self.buffer.getvalue())
+        else:
+            data = self.buffer.getvalue()
```
kyleknap, line 2184: Is there a particular reason we cannot use memoryviews for the other upload if branches here?

Could we instead just have it something like this?
```python
data = self.buffer.getvalue()
if self._sdk_supports_memoryview_for_writes():
     # Add comment on why wrap in memoryview
     data = memoryview(data)
```
Mainly I think this will make the logic a bit more condensed and I think the comment is important to explain that we are wrapping the data so we are not making copies of the bytes when we splice for partitioned uploads.

If we do need the `wb`/`xb` check, maybe let's lower the memoryview logic to the `_stage_block` method and just wrap the `data` provided to it if possible before trying to splice. I'm fine with that approach as well if we want to be more defensive in the approach.

adlfs/spec.py
```diff
@@ -2167,25 +2178,29 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
...
         length = len(data)
         block_id = self._get_block_id(self._block_list)
+        if len(self._block_list) == 0:
+            self._block_list.append(block_id)
```
kyleknap, line 2188: Is there a particular reason we need to add the first block to this list before the first stage block call? In the logic before, it did not seem needed so I'd expect this would hold still and we could just pick it up as part of the `gather()` call.

kyleknap (commented):
Looks good. Just had a couple more comments.

adlfs/spec.py
anjaliratnam-msft: Some of the other tests don't pass if data is a memoryview for the ab mode. I believe the error was with reading the chunk that was added, so I didn't want to change the tests. I can add the memoryview logic to the _stage_block method.

adlfs/spec.py
anjaliratnam-msft: The first block id is always a bunch of zeros, so since the block_list is empty, all the blocks have that same id since they're only added after the tasks run.

adlfs/spec.py
kyleknap: Adding it to stage block sounds good to me. It also limits this optimization to just the code path that needs it.

adlfs/spec.py
```diff
@@ -2153,9 +2154,22 @@ def _get_chunks(self, data):
...
         while start < length:
             end = min(start + self.blocksize, length)
-            yield data[start:end]
+            yield start, end
             start = end
 
+    async def _stage_block(self, data, start, end, block_id, semaphore):
+        if self._sdk_supports_memoryview_for_writes():
+            # Use memoryview to avoid making copies of the bytes when we splice for partitioned uploads
+            data = memoryview(data)
```
kyleknap, line 2163: Let's wrap the data in a memoryview after we've entered the semaphore context so we upload the property that no logic is run until there is availability in the semaphore as we are effectively treating this method as a worker now.

adlfs/spec.py
```diff
@@ -2202,7 +2215,7 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
                 # which is throws an InvalidHeader error from Azure, so instead
                 # of staging a block, we directly upload the empty blob
                 # This isn't actually tested, since Azureite behaves differently.
-                if block_id == self._get_block_id([]) and length == 0 and final:
+                if len(self._block_list) == 0 and length == 0 and final:
```
kyleknap, line 2218: Small nitpick, but we should just be able to do:
```python 
if not self._block_list and length == 0 and final:
```
It's a little bit shorter and more common to check if a list is falsey in order to determine if it is empty.

kyleknap (commented):
Looks good. Just had a couple of more comments.

adlfs/spec.py
```diff
@@ -2169,23 +2183,22 @@ async def _async_upload_chunk(self, final: bool = False, **kwargs):
...
-                    ) as bc:
-                        await bc.stage_block(
-                            block_id=block_id,
-                            data=chunk,
-                            length=len(chunk),
-                        )
-                        self._block_list.append(block_id)
-                        block_id = self._get_block_id(self._block_list)
+                max_concurrency = self.fs.max_concurrency or 1
+                semaphore = asyncio.Semaphore(max_concurrency)
```
martindurant, line 2193: FYI: this effectively makes a "pool", whereas fsspec upstream has a run-coroutines-in-chunks ability. The latter is less efficient but easier to reason about. It may be worthwhile upstreaming this at some point. See a similar idea in https://github.com/fsspec/filesystem_spec/pull/1908 .

martindurant: ( the chunks version: https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L204 )

kyleknap: Upstreaming makes sense to me. We originally did not look upstream in fsspec core to see if there was anything to leverage, but I think generalizing this semaphore approach upstream would definitely be useful to enable concurrency in more places and not need to reduplicate the logic.

Issue #494, opened 2025-02-22
Data-Upload: Timeout problems and chunk size


juergend7lytix:
When uploading large data, I run into a TimeOut-Exception.

I shortened the trace for brevity -- at the end of the issue is the long version.

```python
---------------------------------------------------------------------------
SocketTimeoutError                        Traceback (most recent call last)
File /usr/local/lib/python3.12/site-packages/azure/core/pipeline/transport/_aiohttp.py:306, in AioHttpTransport.send(self, request, stream, proxies, **config)
    305 socket_timeout = aiohttp.ClientTimeout(sock_connect=timeout, sock_read=read_timeout)
--> 306 result = await self.session.request(  # type: ignore
    307     request.method,
    308     request.url,
    309     headers=request.headers,
    310     data=self._get_request_data(request),
    311     timeout=socket_timeout,
    312     allow_redirects=False,
    313     proxy=proxy,
    314     **config,
    315 )
    316 if _is_rest(request):
...

File /usr/local/lib/python3.12/site-packages/aiohttp/streams.py:672, in DataQueue.read(self)
    671 try:
--> 672     await self._waiter
    673 except (asyncio.CancelledError, asyncio.TimeoutError):

SocketTimeoutError: Timeout on reading data from socket

The above exception was the direct cause of the following exception:

ServiceResponseError                      Traceback (most recent call last)
File /usr/local/lib/python3.12/site-packages/adlfs/spec.py:2153, in AzureBlobFile._async_upload_chunk(self, final, **kwargs)
   2150 async with self.container_client.get_blob_client(
   2151     blob=self.blob
   2152 ) as bc:
-> 2153     await bc.stage_block(
   2154         block_id=block_id,
   2155         data=chunk,
   2156         length=len(chunk),
   2157     )
   2158     self._block_list.append(block_id)
...

ServiceResponseError: Timeout on reading data from socket

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)

File /usr/local/lib/python3.12/site-packages/fsspec/spec.py:2227, in AbstractBufferedFile.__exit__(self, *args)
   2226 def __exit__(self, *args):
-> 2227     self.close()
...

File /usr/local/lib/python3.12/site-packages/adlfs/spec.py:2200, in AzureBlobFile._async_upload_chunk(self, final, **kwargs)
   2198                     raise FileExistsError(self.path)
   2199         else:
[...]

PR #509, opened 2025-08-11
updated default block size

anjaliratnam-msft:
Updated the default block size to be 50 MiB so AzureBlobFileSystem and AzureBlobFile are more consistent. This also addresses the [github issue](https://github.com/fsspec/adlfs/issues/494) where the chunk_size of 1GB results in a Timeout Error and also respects the block_size parameter sent in when opening a file. 

I also created this [gist ](https://gist.github.com/anjaliratnam-msft/26e070834411847f6f2cd40d4ec23642)to show that the performance is not impacted negatively when increasing the default from 5 to 50 MiB and performance improves when reducing the chunk size from 1 GB to 50 MiB.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -69,7 +69,7 @@ VERSIONED_BLOB_PROPERTIES = [
     "is_current_version",
 ]
 _ROOT_PATH = "/"
-_DEFAULT_BLOCK_SIZE = 4 * 1024 * 1024
+_DEFAULT_BLOCK_SIZE = 50 * 2**20
 
 _SOCKET_TIMEOUT_DEFAULT = object()
 
@@ -1901,8 +1901,6 @@ class AzureBlobFileSystem(AsyncFileSystem):
 class AzureBlobFile(AbstractBufferedFile):
     """File-like operations on Azure Blobs"""
 
-    DEFAULT_BLOCK_SIZE = 5 * 2**20
-
     def __init__(
         self,
         fs: AzureBlobFileSystem,
@@ -1972,7 +1970,7 @@ class AzureBlobFile(AbstractBufferedFile):
         self.loop = self._get_loop()
         self.container_client = self._get_container_client()
         self.blocksize = (
-            self.DEFAULT_BLOCK_SIZE if block_size in ["default", None] else block_size
+            _DEFAULT_BLOCK_SIZE if block_size in ["default", None] else block_size
         )
         self.loc = 0
         self.autocommit = autocommit
@@ -2146,7 +2144,7 @@ class AzureBlobFile(AbstractBufferedFile):
 
     _initiate_upload = sync_wrapper(_async_initiate_upload)
 
-    def _get_chunks(self, data, chunk_size=1024**3):  # Keeping the chunk size as 1 GB
+    def _get_chunks(self, data, chunk_size):
         start = 0
         length = len(data)
         while start < length:
@@ -2173,7 +2171,7 @@ class AzureBlobFile(AbstractBufferedFile):
             commit_kw["headers"] = {"If-None-Match": "*"}
         if self.mode in {"wb", "xb"}:
             try:
-                for chunk in self._get_chunks(data):
+                for chunk in self._get_chunks(data, self.blocksize):
                     async with self.container_client.get_blob_client(
                         blob=self.blob
                     ) as bc:
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -177,8 +177,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         The credentials with which to authenticate.  Optional if the account URL already has a SAS token.
         Can include an instance of TokenCredential class from azure.identity.aio.
     blocksize: int
-        The block size to use for download/upload operations. Defaults to hardcoded value of
-        ``BlockBlobService.MAX_BLOCK_SIZE``
+        The block size to use for download/upload operations. Defaults to 50 MiB
     client_id: str
         Client ID to use when authenticating using an AD Service Principal client/secret.
     client_secret: str
@@ -1879,6 +1878,8 @@ class AzureBlobFileSystem(AsyncFileSystem):
             is versioning aware and blob versioning is enabled on the releveant container.
         """
         logger.debug(f"_open:  {path}")
+        if block_size is None:
+            block_size = self.blocksize
         if not self.version_aware and version_id:
             raise ValueError(
                 "version_id cannot be specified if the filesystem "
@@ -1900,6 +1901,8 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
 class AzureBlobFile(AbstractBufferedFile):
     """File-like operations on Azure Blobs"""
+
+    DEFAULT_BLOCK_SIZE = _DEFAULT_BLOCK_SIZE
 
     def __init__(
         self,
@@ -1970,7 +1973,7 @@ class AzureBlobFile(AbstractBufferedFile):
         self.loop = self._get_loop()
         self.container_client = self._get_container_client()
         self.blocksize = (
-            _DEFAULT_BLOCK_SIZE if block_size in ["default", None] else block_size
+            self.DEFAULT_BLOCK_SIZE if block_size in ["default", None] else block_size
         )
         self.loc = 0
         self.autocommit = autocommit
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1862,7 +1862,8 @@ class AzureBlobFileSystem(AsyncFileSystem):
             What mode to open the file in - defaults to "rb"
 
         block_size: int
-            Size per block for multi-part downloads.
+            Size per block for multi-part uploads and downloads. This overrides the block_size parameter
+            and if not provided, defaults to the filesystem blocksize.
 
         autocommit: bool
             Whether or not to write to the destination directly
@@ -2147,11 +2148,11 @@ class AzureBlobFile(AbstractBufferedFile):
 
     _initiate_upload = sync_wrapper(_async_initiate_upload)
 
-    def _get_chunks(self, data, chunk_size):
+    def _get_chunks(self, data):
         start = 0
         length = len(data)
         while start < length:
-            end = min(start + chunk_size, length)
+            end = min(start + self.blocksize, length)
             yield data[start:end]
             start = end
 
@@ -2174,7 +2175,7 @@ class AzureBlobFile(AbstractBufferedFile):
             commit_kw["headers"] = {"If-None-Match": "*"}
         if self.mode in {"wb", "xb"}:
             try:
-                for chunk in self._get_chunks(data, self.blocksize):
+                for chunk in self._get_chunks(data):
                     async with self.container_client.get_blob_client(
                         blob=self.blob
                     ) as bc:
--- a/adlfs/tests/test_spec.py
+++ b/adlfs/tests/test_spec.py
@@ -2050,6 +2050,7 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
 
 def test_number_of_blocks(storage, mocker):
     from azure.storage.blob.aio import BlobClient
+
     blocksize = 5 * 2**20
     fs = AzureBlobFileSystem(
         account_name=storage.account_name,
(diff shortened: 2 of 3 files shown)


adlfs/spec.py
```diff
@@ -2146,7 +2147,7 @@ async def _async_initiate_upload(self, **kwargs):
 
     _initiate_upload = sync_wrapper(_async_initiate_upload)
 
-    def _get_chunks(self, data, chunk_size=1024**3):  # Keeping the chunk size as 1 GB
+    def _get_chunks(self, data, chunk_size):
```
kyleknap, line 2150: Looking through this more, it probably makes sense to just remove the `chunk_size` parameter from this helper method all together and just switch the sole reference to chunk_size with the instance property `self.blocksize`. Mainly, we are not really overriding this value now and the block size is already set in the constructor. So it makes it simpler.

adlfs/spec.py
```diff
@@ -69,7 +69,7 @@
     "is_current_version",
 ]
 _ROOT_PATH = "/"
-_DEFAULT_BLOCK_SIZE = 4 * 1024 * 1024
+_DEFAULT_BLOCK_SIZE = 50 * 2**20
```
kyleknap, line 72: We should also make sure to update the changelog in this PR. I'm thinking we can just use the three bullet points from this comment I made: https://github.com/fsspec/adlfs/pull/509#issuecomment-3180475337 and make the wording a bit more succinct.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+        with patch(
+            "azure.storage.blob.aio.BlobClient.stage_block", autospec=True
+        ) as mock_stage_block:
+            f.write(content)
+            expected_blocks = math.ceil(len(content) / blocksize)
+            actual_blocks = mock_stage_block.call_count
+            assert actual_blocks == expected_blocks
+
+
+def test_block_size(storage):
```
kyleknap, line 2075: It might be worth parameterizing this one to test out the different permutations of setting/omitting block sizes.  To do this, we parameterize off of the input block size for both the file system and the `fs.open()` call and include the expected block size for both the file system and file object. We'd then add cases such as (feel free to adjust this or add more):
1. Assert defaults when block size is not set for either the file system or file object
2. Assert file system block size propagates to the file-like object
3. Assert that we can override the `block_size` for the `fs.open()` call

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+    fs = AzureBlobFileSystem(
+        account_name=storage.account_name,
+        connection_string=CONN_STR,
+        blocksize=5 * 2**20,
```
kyleknap, line 2079: When implementing cases where we override the block size. It may be worth using values that are unlikely/odd defaults (e.g. `7 * 2 ** 20`) to make it more clear this is an override value as opposed to a possible default.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+    )
+
+    with fs.open("data/root/a/file.txt", "wb") as f:
+        assert f.blocksize == 5 * 2**20
```
kyleknap, line 2083: We should probably add an additional parameterized test case where we set the `block_size` directly using the `AzureBlobFile` constructor and have cases that:
1. Assert the default block size for both when `block_size` is ommitted and `block_size` is `None`.
2. Assert block size is **not** inherited from the passed in `fs`. This one may make sense to implement as a separate test method in case the implementation is not clean as part of parameterization
3. Assert that we can override `block_size`

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+def test_number_of_blocks(storage, mocker, blocksize):
+
+    fs = AzureBlobFileSystem(
+        account_name=storage.account_name,
+        connection_string=CONN_STR,
+        blocksize=blocksize,
+    )
+
+    content = b"1" * (blocksize * 2 + 1)
+    with fs.open("data/root/a/file.txt", "wb", blocksize=blocksize) as f:
```
kyleknap, line 2062: Let's remove `blocksize` from this call here as I'm not sure if this will affect anything since the argument used is `block_size` and it should suffice to set it in the file system.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+        mocker.patch(
+            "azure.storage.blob.aio.BlobClient.commit_block_list", autospec=True
+        )
+        with patch(
+            "azure.storage.blob.aio.BlobClient.stage_block", autospec=True
+        ) as mock_stage_block:
```
kyleknap, line 2068: For these patch statements, is there a particular reason why:

1. `mocker.patch` is not being used for both?
2. We are not following the pattern from other test cases where we import the `BlobClient` first e.g., [here](https://github.com/anjaliratnam-msft/adlfs/blob/2559f6e2f5d89e94f07fcb51a616f7a10ad26ce8/adlfs/tests/test_spec.py#L1950) and patching it directly?

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+            f.write(content)
+            expected_blocks = math.ceil(len(content) / blocksize)
+            actual_blocks = mock_stage_block.call_count
+            assert actual_blocks == expected_blocks
```
kyleknap, line 2072: Since we are patching the commit block list call as well, it may be worth to also assert the number of blocks in that call also match the expected number of blocks.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
         with fs.open("data/afile", "xb") as f:
             pass
     assert fs.cat_file("data/afile") == b"data"
+
+
+@pytest.mark.parametrize("blocksize", [5 * 2**20, 50 * 2**20, 100 * 2**20])
```
kyleknap, line 2052: I think we should be fine just making this a non-parameterized test. Not sure if there is much value in setting different block sizes. I'd say we just set the blocksize to somewhere in single MiB's and `write()` data that requires several blocks and maybe also make sure the last block does not completely fit completely into a block (e.g., is less than blocksize) to make sure the logic handles sizes that do not fit on block boundaries.

This will hopefully help simplify the scaffolding for the case and also take less time to run.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2047,37 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+    with fs.open("data/root/a/file.txt", "wb", blocksize=blocksize) as f:
+        mocker.patch(
+            "azure.storage.blob.aio.BlobClient.commit_block_list", autospec=True
+        )
+        with patch(
+            "azure.storage.blob.aio.BlobClient.stage_block", autospec=True
+        ) as mock_stage_block:
+            f.write(content)
+            expected_blocks = math.ceil(len(content) / blocksize)
+            actual_blocks = mock_stage_block.call_count
```
kyleknap, line 2071: It would also probably be worth asserting the actual data sizes used in each `stage_block` call.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2046,82 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+        assert sum(block_lengths) == len(content)
+
+    assert len(mock_commit_block_list.call_args.kwargs["block_list"]) == expected_blocks
+
+
+@pytest.mark.parametrize(
+    "filesystem_blocksize, file_blocksize, expected_blocksize",
+    [
+        (None, None, 50 * 2**20),
+        (50 * 2**20, None, 50 * 2**20),
```
kyleknap, line 2081: For this case, it might make sense to make it a different value like that `7 * 2 ** 20` value to disambiguate from the default value.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2046,82 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+        (None, 5 * 2**20, 5 * 2**20),
+        (50 * 2**20, 7 * 2**20, 7 * 2**20),
```
kyleknap, line 2083: Same thing here, instead of using `50 * 2 ** 20` let's use a non-default value like 40 * 2 ** 20 to disambiguate from any defaults.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2046,82 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+        block_lengths = [
+            call.kwargs["length"] for call in mock_stage_block.call_args_list
+        ]
+        assert sum(block_lengths) == len(content)
+
+    assert len(mock_commit_block_list.call_args.kwargs["block_list"]) == expected_blocks
+
+
+@pytest.mark.parametrize(
+    "filesystem_blocksize, file_blocksize, expected_blocksize",
```
kyleknap, line 2078: It's also probably worth adding an `expected_filesystem_blocksize` to make the blocksize are expected at the filesystem level as well in this test case.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2046,82 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+    fs = AzureBlobFileSystem(
+        account_name=storage.account_name, connection_string=CONN_STR
+    )
+    f = AzureBlobFile(
+        fs,
+        "data/root/a/file.txt",
+        block_size=file_blocksize,
+    )
+    assert f.blocksize == expected_blocksize
+    assert fs.blocksize == 50 * 2**20
```
kyleknap, line 2114: We can probably remove this assertion assuming we add the `expected_filesystem_blocksize` parameter in the other parameterized test case.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2046,82 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
...
+
+
+def test_override_blocksize(storage):
```
kyleknap, line 2117: Ah so for this test case I was thinking, we'd:
1. Set the `blocksize` on the `AzureBlobFileSystem`
2. Assert that the `AzureBlobFile` without setting `block_size` does not inherit the the blocksize from 1. So after instantiation, we would just assert it is `50 * 2 ** 20`, which should be its default.
3. Update the test name to better reflect that we are testing this code path does not inherit from the file system.

adlfs/tests/test_spec.py
```diff
@@ -2045,3 +2046,82 @@ def test_open_file_x(storage: azure.storage.blob.BlobServiceClient, tmpdir):
         with fs.open("data/afile", "xb") as f:
             pass
     assert fs.cat_file("data/afile") == b"data"
+
+
+def test_number_of_blocks(storage, mocker):
```
kyleknap, line 2051: Let's update this test name to be `test_uses_block_size_for_partitioned_uploads` in order to be a bit more self-descriptive of what we are trying to test.

kyleknap (commented):
Looks good! Just had some really small comments on the test and should be set from a code change perspective.

kyleknap (approved):
Looks great! Thank you for putting together these [performance results](https://gist.github.com/anjaliratnam-msft/26e070834411847f6f2cd40d4ec23642). This is really helpful in making decisions on this.

I think it is awesome to see that with this PR alone, we will be reducing timeout/network errors by sending smaller upload payloads (50 MiB instead of possibly 1 GiB) and we will greatly be increasing the speed for small sequential writes (e.g. writing 1KB at a time scenario) as we will be reducing the number of PUTs being made.

Furthermore, it is great to see when we then we add on this [PR](https://github.com/fsspec/adlfs/pull/508) that adds concurrency to partitioned uploads, it improves the speed of large writes (e.g., writing a large tensor in a PyTorch model) by 2x - 3x by default and can reach up to 5x improvement if you further tune blocksize and concurrency based on your workload. With both of these PRs, it really puts write performance on par with the read performance of adlfs now. 

@martindurant this all should be ready to review. @anjaliratnam-msft includes the rationale in the [perf results gist](https://gist.github.com/anjaliratnam-msft/26e070834411847f6f2cd40d4ec23642) but we decided to stick with the approach of 50 MiB block size. We did continue to see when concurrency is enabled and saturated, 5 MiB block sizes squeezed out more speed than the 50 MiB sizes. However, the tradeoff in further limiting default blob sizes and performance degradation when concurrency is disabled outweighed these possible gains.

Let us know what you think!

PR #514, opened 2025-10-04
Revise README for Azure Blob and Data Lake Storage

tjvishnu:
-Removed references to ADLS Gen1, which is retired.  
-Added information on connecting to Blobs as well as ADLS Gen2

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,9 @@ or
 
 `conda install -c conda-forge adlfs`
 
-The `az://` and `abfs://` protocols are included in fsspec's known_implementations registry 
-in fsspec > 0.6.1, otherwise users must explicitly inform fsspec about the supported adlfs protocols.
+The `az://` and `abfs://` protocols are included in fsspec's known_implementations registry.
 
-To connect to Blobs or Azure Data Lake Storage (ADLS) Gen2 filesystem you can use the protocol `abfs` or `az`:
+To connect to Azure Blob Storage or Azure Data Lake Storage (ADLS) Gen2 filesystem you can use the protocol `abfs` or `az`:
 
 ```python
 import dask.dataframe as dd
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,7 +66,7 @@ The `storage_options` can be instantiated with a variety of keyword arguments de
    anonymous access will not be attempted. Otherwise the value for `anon` resolves to True.
 - `location_mode`: valid values are "primary" or "secondary" and apply to RA-GRS accounts
 
-For more argument details see all arguments for [`AzureBlobFileSystem` here](https://github.com/fsspec/adlfs/blob/f15c37a43afd87a04f01b61cd90294dd57181e1d/adlfs/spec.py#L328) and [`AzureDatalakeFileSystem` here](https://github.com/fsspec/adlfs/blob/f15c37a43afd87a04f01b61cd90294dd57181e1d/adlfs/spec.py#L69).
+For more argument details see all arguments for [`AzureBlobFileSystem` here](https://github.com/fsspec/adlfs/blob/f15c37a43afd87a04f01b61cd90294dd57181e1d/adlfs/spec.py#L328)
```
kyleknap, line 69: Instead of linking to the code, let's instead link to the rendered HTML docs: https://fsspec.github.io/adlfs/api/#adlfs.AzureBlobFileSystem. Mainly, there is a hardcoded commit SHA here so if arguments are added in the future, this link will drift away from what is in main.

README.md
```diff
@@ -102,3 +87,6 @@ The filesystem can be instantiated for different use cases based on a variety of
 The `AzureBlobFileSystem` accepts [all of the Async BlobServiceClient arguments](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python).
 
 By default, write operations create BlockBlobs in Azure, which, once written can not be appended. It is possible to create an AppendBlob using `mode="ab"` when creating and operating on blobs. Currently, AppendBlobs are not available if hierarchical namespaces are enabled.
+
+### Older versions
+ADLS Gen1 filesystem has officially been [retired](https://learn.microsoft.com/en-us/lifecycle/products/azure-data-lake-storage-gen1)). Hence the older versions of this package, which was designed to connect to ADLS Gen1 is obsolete.
```
kyleknap, line 92: A couple of suggestions on this section:
* It looks like there is an extra `)` that we should remove at the end of the link.
* Maybe we can remove this second sentence and just add on: "and support in `adlfs` is obsolete"? Mainly, the AzureBlobFileSystem has been around for several years so technically these older versions should work just fine if customers were using `az` or `abfs`.

kyleknap (commented):
Looks good! Just had a few small comments.

PR #517, opened 2025-11-04
Move metadata fetching to get_metadata method

benjaminkaplan:
Refactor metadata retrieval into a separate method.  No reason to fetch metadata synchronously when creating the `AzureBlobFile` object.  Still enables metadata fetching using the new `get_metadata` method.

This saves about ~50ms per file open.

Generally synchronous calls are avoided without the user specifically requesting them. Therefore there is no reason to perform a sync call every time an object is created.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -2056,14 +2056,16 @@ class AzureBlobFile(AbstractBufferedFile):
             self.offset = None
             self.forced = False
             self.location = None
+
     def get_metadata(self):
-            return self.metadata or sync(self.loop,
-                                         get_blob_metadata,
-                                         self.container_client,
-                                         self.blob,
-                                         version_id=self.version_id,
-                                        )
-            
+        return self.metadata or sync(
+            self.loop,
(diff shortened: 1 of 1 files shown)


benjaminkaplan:
I submitted a fix for a typo in my previous commit on this PR. Can you please run the tests again

benjaminkaplan:
I see the tests are failing but with no explanation. It just says `Process finished with exit code 1`

Is there a way to see why this is failing?

anjaliratnam-msft:
@benjaminkaplan 
This makes sense to not fetch the metadata every time for reads, however, if someone were to access the metadata parameter, it would no longer be populated which could cause problems. Instead maybe it would make more sense to lazily load it and only populate the metadata when it is actually accessed? We should also probably add a test to make sure this behavior is as expected. Open to other ideas as well though! 

For the failing tests, it's due to linting errors, if you click on the check you should be able to see the exact error. Running `pre-commit run --all-files` should take care of this!

PR #524, opened 2025-12-04
capture version_id after upload

shcheklein:
Similar to https://github.com/fsspec/gcsfs/pull/718

To be able to refer to a file properly (in case multiple uploads happening simultaneously to the same path), we want to grab generation from the response. S3 already does this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Unreleased
 ----------
 - Added `**kwargs` to `AzureBlobFileSystem.exists()`
+- Populate `AzureBlobFile.version_id` on write when `version_aware` is enabled.
 
 2026.2.0
 --------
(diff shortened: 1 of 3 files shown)


shcheklein:
@martindurant do you know by chance who is maintaining the project and could release this change?

kyleknap:
@shcheklein thanks for contribution! Populating the updated `version_id` after a completed upload seems reasonable to me especially since `s3fs` is already doing this and you have a similar change out for `gcsfs`.

I'll have more cycles next week to take a deeper review but some initial things we probably want to address now are:
1. Let's populate the `version_id` only if the upstream file system is `version_aware`. Looks like `s3fs` has [similar logic](https://github.com/fsspec/s3fs/blob/65f394575b9667f33b59473dc28a8f1cf6708745/s3fs/core.py#L2528-L2529)
2. It would be great if we can get test cases added to make sure we the `version_id` is getting set for the different upload branches where it is being set now for version aware filesystems. It would also be good to have a test case too where we don't set the `version_id` if the filesystem is not version aware.
3. Add a small changelog entry [here](https://github.com/fsspec/adlfs/blob/main/CHANGELOG.md#unreleased) for the enhancement.

kyleknap:
Thanks @shcheklein! I agree mocks suffice in this case. I'll take a deeper look at the PR this week.

kyleknap (commented):
This looks good! I pulled down the PR and tried it out and it was working well. I just had one question related to append blobs that I ran into.

PR #526, opened 2026-01-02
Remove upper bound for azure-datalake-store dependency

hutch3232:
I noticed there is a 1.x.x release (June, 2025) of this dependency but this upper bound prevents it's usage. However, it is being tested against already (the latest.txt requirements are uncapped).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ requires-python = ">=3.10"
 dynamic = ["version"]
 dependencies = [
     "azure-core>=1.28.0,<2.0.0",
-    "azure-datalake-store>=0.0.53,<0.1",
+    "azure-datalake-store>=0.0.53",
     "azure-identity",
     "azure-storage-blob>=12.17.0",
     "fsspec>=2023.12.0",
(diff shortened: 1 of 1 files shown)


potiuk:
> The diff looks good, but before merging it, I'd like to pull it down and try to vet if there are any negative impact of the upgrade especially since there are no tests for it. I'm hoping to do that in the next few days.

Maybe we can help with that - if you have a version of azure-data-lake-store (say alpha/beta/rc) that we can test, we can run  it through our (Apache Airflow) testing suite for our provider - and maybe other users can be asked for it as well - we do not have a full coverage of it of course, but when you combine inputs from multiple users, your own testing might be limited

kyleknap:
@hutch3232 Got it. Thanks for the context. I still think running it through your test suite would be helpful. Mainly curious on whether you had an end to end set up. I plan to try to get a working setup when I pull down the PR and see how far I get.

kyleknap:
We just merged the PR to remove ADLS Gen 1 support (including the `azure-datalake-store` dependency): https://github.com/fsspec/adlfs/pull/545. It will be available in the next release of adlfs. Closing this PR as adlfs no longer depends on the ADLS Gen 1 `azure-datalake-store` library.

PR #528, opened 2026-01-14
Move metadata evaluation to a lazy property

nateprewitt:
This PR makes the metadata property on `AzureBlobFile` lazy-loaded for read mode, addressing the performance issue raised in #517. Previously, metadata was fetched synchronously every time a file was opened (~50ms overhead), regardless of use. Now it's only fetched on first access of the .metadata property. This approach preserves backwards compatibility while reducing average overhead.

The one assumption being made is end-users aren't manually assigning `metadata` from their code. That _seems_ like an unlikely  edge case. If we're concerned about that though, I can add a setter as well. I just don't want to encourage that behavior if we don't consider it supported/intended.

### Changes
* Removes eager `get_blob_metadata` call from `__init__` for read mode.
* Adds `@property` decorator to fetch and cache metadata on first access.
* Includes tests verifying lazy-loading behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 Unreleased
 ----------
 
+* Lazy-load `metadata` property on `AzureBlobFile` to avoid fetching metadata on file open
+  when it is not accessed.
 * Respect `AzureBlobFileSystem.protocol` tuple when removing protocols from fully-qualified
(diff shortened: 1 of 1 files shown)


anjaliratnam-msft (commented):
This looks good to me, I just had one small comment!

PR #530, opened 2026-01-23
Added deprecation warning for anonymous access

anjaliratnam-msft:
Added a deprecation warning for changing the anonymous access default to False. The message will only show up if users do not provide a credential or explicitly set `anon`.

To continue using anonymous access, users should pass in `anon=True`. e.g.:
`AzureBlobFileSystem(account_name=<your-account-name>,...,anon=True,)`


adlfs/tests/test_spec.py
```diff
@@ -2372,3 +2373,48 @@ def test_metadata_setter(storage, mocker):
...
+        (None, None, None, None, None, True),
+        (None, None, "credential", None, None, False),
+        (None, None, None, "sas_token", None, False),
+        (None, None, None, None, "account_key", False),
+        (None, "true", None, None, None, False),
+        (None, "false", None, None, None, False),
+        (None, "true", "credential", None, None, False),
+        ("true", None, None, None, None, False),
+        ("false", None, None, None, None, False),
+        ("true", None, "credential", None, None, False),
```
nateprewitt: anon is defined as an [optional bool](https://github.com/fsspec/adlfs/blob/18cbbed88eac54885e08d194902df777ba2456c0/adlfs/spec.py#L278) so we should try to keep tests reflecting the public API.
```suggestion
        (True, None, None, None, None, False),
        (False, None, None, None, None, False),
        (False, None, "credential", None, None, False),
```

adlfs/tests/test_spec.py
```diff
@@ -2372,3 +2373,48 @@ def test_metadata_setter(storage, mocker):
...
+    ],
+)
+def test_anon_default_warning(
+    storage, anon, env_value, credential, sas_token, account_key, warning
+):
+    env_var = {} if env_value is None else {"AZURE_STORAGE_ANON": env_value}
+    with mock.patch.dict(os.environ, env_var):
+        if warning:
```
nateprewitt: This is more high-level testing advice for Python, so you can do with it what you want. Generally, it's good to avoid conditionals inside tests. Once you introduce branching logic, it's not uncommon to silently break things while the tests continue to report success. 

`pytest.mark.parameterize` is great for simplifying code, but there's also cases where you may want to make distinctions between what you're testing (positive cases, negative cases, env vs not, etc).

You might consider two tests here. We can also get a little clever with the testing inputs if you want, but not required at all.

`test_anon_warning` (positive test for case 1, ensure the warning is raised where we expect)
`test_no_anon_warning` (negative tests for cases 2-10)

kyleknap:
Yeah we definitely do not want to warn if connection string is provided; connection strings count as one of the ways to provide credentials and avoid defaulting to anonymous credentials.

adlfs/tests/test_spec.py
```diff
@@ -2372,3 +2373,48 @@ def test_metadata_setter(storage, mocker):
...
+
+@pytest.mark.parametrize(
+    "env_vars,storage_options",
+    [
+        (None, {"credential": "credential"}),
+        (None, {"sas_token": "sas_token"}),
+        (None, {"account_key": "account_key"}),
+        ({"AZURE_STORAGE_ANON": "true"}, {}),
+        ({"AZURE_STORAGE_ANON": "false"}, {}),
+        ({"AZURE_STORAGE_ANON": "true", "credential": "credential"}, {}),
```
kyleknap: Is the `credential` for this case supposed to be in the `storage_options` and not the `env_vars`?

PR #536, opened 2026-02-13
Add **kwargs to exists()

### LINKED ISSUE #469: exists() is missing **kwargs
The spec says that `exists()` takes in `**kwargs`: https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/spec.html#AbstractFileSystem.exists

but https://github.com/fsspec/adlfs/blob/main/adlfs/spec.py#L1368 does not.

anjaliratnam-msft:
Added `**kwargs` to `exists()` to be consistent with the fsspec implementation. This resolves #469.


martindurant:
You may choose to add a test exercising the kwargs (which have no effect)

PR #537, opened 2026-02-18
Clarify async credential requirement and update examples

nateprewitt:
`adlfs` uses the async Azure SDK under the hood, which requires async credentials. A common source of issues is users passing synchronous credentials (e.g. `azure.identity.DefaultAzureCredential`) instead of their async counterparts (`azure.identity.aio.DefaultAzureCredential`). This was partly due to inaccurate documentation in the project which should now be fixed.

For almost all users, the preferred option is to use `anon=False` (soon to be default) when instantiating your file system. This allows the SDK to auto-detect the right credentials. If you must supply credentials directly, make sure you're using the async version.

This PR:
- Emits a `FutureWarning` when synchronous credentials are detected
- Adds a `CredentialType` alias for accurate type checking
- Updates docs and README to recommend `anon=False` over passing credential objects directly, and corrects references to point to `azure.identity.aio`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Unreleased
 ----------
 - Added `**kwargs` to `AzureBlobFileSystem.exists()`
+- Populate `AzureBlobFile.version_id` on write when `version_aware` is enabled.
 - Fixed issue where unawaitable Credential types were incorrectly awaited (#431)
 
 2026.2.0
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -2003,9 +2003,10 @@ class AzureBlobFile(AbstractBufferedFile):
             by `cache_type`.
 
         version_id : str
-            Optional version to read the file at.  If not specified this will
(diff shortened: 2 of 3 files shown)


anjaliratnam-msft (commented):
This looks good to me!

adlfs/utils.py
```diff
@@ -81,5 +81,7 @@ async def close_credential(file_obj):
     Implements asynchronous closure of credentials for
     AzureBlobFile objects
     """
-    if not isinstance(file_obj.credential, (type(None), str)):
-        await file_obj.credential.close()
+    if not isinstance(file_obj.credential, (type(None), str, dict)):
+        result = file_obj.credential.close()
+        if result is not None:
```
kyleknap: Would it be possible to check if this is awaitable either using `inspect.isawaitable()` or instance check on `collections.abc.Awaitable`? Mainly synchronous `close()` should not be returning anything but might be safer in the future if for whatever reason it does.

PR #538, opened 2026-02-18
Add automated type checking

nateprewitt:
This is a draft PR to serve as a talking point on how we'd enable automated type checking for current and future PRs. That will help enforce the existing typing and make sure we're keeping types up to date.

This PR touches a number of places in the code base so I've split it into 3 distinct commits.

* Enabling automated type checking with pyright, for now this is non-blocking for merge.
* Fixing existing bugs/inaccurate typing in the code base.
* Migrating remaining typing to [PEP 585](https://peps.python.org/pep-0585/) and [PEP 604](https://peps.python.org/pep-0604/).

There are some cases where we're adding inline ignores so I'm going to comment on those specifically with context. We can discuss the tradeoffs there. We'll also need to get the other outstanding type PRs merged before this is ready for final review.

### DIFF
--- a/adlfs/gen1.py
+++ b/adlfs/gen1.py
@@ -186,7 +186,7 @@ class AzureDatalakeFileSystem(AbstractFileSystem):
 
     def __setstate__(self, state):
         logger.debug("De-serialize with state: %s", state)
-        self.__dict__.update(state)  # type: ignore[reportAttributeAccessIssue]
+        self.__dict__.update(state)
         self.do_connect()
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ namespaces = false
 
 [tool.pyright]
 include = ["adlfs"]
-exclude = ["adlfs/tests", "build"]
+exclude = ["adlfs/tests", "adlfs/gen1.py", "build"]
 pythonVersion = "3.10"
 typeCheckingMode = "basic"
 


adlfs/gen1.py
```diff
@@ -186,7 +186,7 @@ def __getstate__(self):
 
     def __setstate__(self, state):
         logger.debug("De-serialize with state: %s", state)
-        self.__dict__.update(state)
+        self.__dict__.update(state)  # type: ignore[reportAttributeAccessIssue]
```
nateprewitt: This is an interaction with pyright and inheriting from a 3p untyped library. While this should be a generic `Mapping`/`dict` as an instance of the class, pyright is inferring it to be `MappingProxyType[str, Any]` which isn't guaranteed to have `update`.

I believe we either need to get this addressed at the pyright level or with typing in fsspec. Neither of those are likely to happen soon, so I've commented this out.

I also know we may remove this gen1.py file at some point but this same issue occurs elsewhere in the changes.

adlfs/spec.py
```diff
@@ -407,7 +409,7 @@ def _strip_protocol(cls, path: str):
             Returns a path without the protocol
         """
         if isinstance(path, list):
-            return [cls._strip_protocol(p) for p in path]
+            return [cls._strip_protocol(p) for p in path]  # type: ignore[return-value]
```
nateprewitt: We already explicitly require `path` to be a `str`. The upstream fsspec version of this also only uses a `str` but has this code path in the event we somehow get a list. This block keeps parity with upstream but isn't possible through the current public API for adlfs.

We can introduce a change that records a return type of `str | list[str]` but it significantly complicates the typing story without providing more accuracy.

adlfs/spec.py
```diff
@@ -473,6 +475,16 @@ def _get_credential_from_service_principal(self):
...
         """
+        if (
+            self.tenant_id is None
+            or self.client_id is None
+            or self.client_secret is None
+        ):
+            raise ValueError(
+                "tenant_id, client_id, and client_secret must all be provided "
+                "when authenticating with a service principal."
+            )
```
nateprewitt: This is an existing failure mode. You can pass just `self.client_id` to trigger this credential path, but it will always fail if all 3 values aren't set. This enforces that requirement on our side rather than getting various errors from the SDK that are more difficult to debug.

adlfs/spec.py
```diff
@@ -1635,6 +1647,12 @@ async def _url(
             account_name = self.account_name
             account_key = self.account_key
 
+        if account_name is None:
+            raise ValueError(
+                "account_name is required to generate a SAS URL. "
+                "Provide account_name or include AccountName in the connection string."
+            )
```
nateprewitt: Same deal about downstream requirements. If account_name is None, the string concatenation blows up because you can't add None to a string. This will raise a clearer error.

adlfs/spec.py
```diff
@@ -2127,9 +2146,9 @@ def connect_client(self):
         """
         try:
             if hasattr(self.fs, "account_host"):
-                self.fs.account_url: str = f"https://{self.fs.account_host}"
+                self.fs.account_url = f"https://{self.fs.account_host}"
```
nateprewitt: These aren't valid type definitions because they're modifying `self.fs` not a member of the class directly. The correct typing already lives on the `fs` implementation.

.github/workflows/tests.yml
```diff
@@ -60,3 +60,27 @@ jobs:
       with:
         file: ./coverage.xml
         fail_ci_if_error: false
+
+  type-check:
+    runs-on: "ubuntu-22.04"
+    continue-on-error: true
```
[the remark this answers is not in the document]
nateprewitt: No we can definitely add them there. This was originally written to get testing bootstrapped on a machine other than my local one but we should make sure it's easily reproducible everywhere.

The continue-on-error was to not block PRs just because there was a typing issue. Whether we want to keep that, I'm indifferent on. What I wanted to avoid was people adding #ignore or making the types potentially worse to get around the typechecker.

adlfs/spec.py
```diff
@@ -386,13 +386,15 @@ def __init__(
             weakref.finalize(self, sync, self.loop, close_credential, self)
 
         if max_concurrency is None:
-            batch_size = _get_batch_size()
+            batch_size: int = _get_batch_size()  # type: ignore[assignment]
             if batch_size > 0:
                 max_concurrency = batch_size
+            else:
+                max_concurrency = 1
```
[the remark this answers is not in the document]
nateprewitt: I have a [PR](https://github.com/Azure/azure-sdk-for-python/pull/45598) open with the Azure SDK to hopefully help fix this issue. The APIs are inconsistently typed, some allow None, others don't. A different subset actually works with None (this is one of them).

Once we get that aligned, we should be fine to remove this piece.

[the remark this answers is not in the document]
nateprewitt: We got 3 PRs merged upstream. Once the next Python SDK release goes out, we should be able to remove this.

Issue #461, opened 2024-02-15
`fs.info()` and `fs.ls(detail=True)` return different etag formats


pmrowla:
I'm not sure this is something that needs to be addressed in adlfs, but I think it's worth noting that etags returned by `fs.info()` and `fs.ls(detail=True)` have different formats. When using `fs.info()` the etag will be quoted (wrapped in `"` quotes) but when using `fs.ls()` the etag will not be quoted. This means that the caller has to correct the quote format before being able to compare etags to see whether a file has been modified.

With adlfs 2024.2.0:
```python
>>> from adlfs import AzureBlobFileSystem
>>> fs = AzureBlobFileSystem()
>>> fs.info("az://test-deletion/2024.2.0/foo", refresh=True)["etag"]
'"0x8DC2DFC9E520378"'
>>> [(f["name"], f["etag"]) for f in fs.ls("az://test-deletion/2024.2.0/", detail=True, refresh=True)]
[('test-deletion/2024.2.0/foo', '0x8DC2DFC9E520378')]
```

I did some investigation and it looks like this is due to differences in what the azure API returns for different calls. When using `BlobClient.get_blob_properties()` the returned etag property is wrapped in `"` quotes. However, when using `BlobContainer.walk_blobs()` the etag property for iterated blob properties is not wrapped in quotes.

wonwuakpa-msft:
Hey @pmrowla, we have included a fix for this in the [2026.5.0 adlfs release.](https://github.com/fsspec/adlfs/releases/tag/2026.5.0) `info()` and `ls()` have consistent etag formatting wrapped in `"`

PR #544, opened 2026-04-21
Normalize Etag format

wonwuakpa-msft:
## Description 

This fixes a bug with the Etag blob property where `info()` and `ls()` were returning different formats. This [GitHub issue](https://github.com/fsspec/adlfs/issues/461) reproduces said bug. This is due to API differences when using `BlobClient.get_blob_properties()` vs  `BlobContainer.walk_blobs()`. 
This PR introduces a standardization to `_details()` by ensuring the Etag property return type is wrapped with double quotes.  S3fs formats Etags with double quotes [here](https://github.com/fsspec/s3fs/blob/731e1250bcd4f682e1ccce03b01641910e7646fc/s3fs/core.py#L1741)

## Test
Added `test_etag_normalized_form` to test both calls property is equivalent and there is no double quoting


adlfs/spec.py
```diff
@@ -897,6 +897,12 @@ async def _details(
                 for key in FORWARDED_BLOB_PROPERTIES
                 if content.has_key(key)  # NOQA
             }
+
+            # Normalize etag to always return a string with quotes for consistency
+            if data.get("etag") and data["etag"] is not None:
```
nateprewitt: I think we're checking the same thing twice here, so we may be able to simplify. `get()` will give us back `None` if it doesn't exist so we can just check the value there. These are a couple alternatives we can use depending on how we want to mutate the variable. More on that below.

```suggestion
            if data.get("etag") is not None:
```
or
```suggestion
            if (etag_val := data.get("etag")) is not None:
```

If we're concerned about the server sending us back an empty string (that shouldn't happen?), we can change our default. This validates the value is "truthy" so we're not quoting an empty string. I'm not actually sure what the desired behavior is here though, I don't think that'd ever be a valid etag.

```suggestion
            if (etag_val := data.get("etag", "")):
```

adlfs/spec.py
```diff
@@ -897,6 +897,12 @@ async def _details(
...
+                etagVal = data["etag"]
```
nateprewitt: Python has a lot of semantics built into casing for variables. [PEP 8](https://peps.python.org/pep-0008/#prescriptive-naming-conventions) codifies most of those and is a really useful reference when in doubt. Function scope variable names should be snake_case (i.e. `etag_val` in this case).

adlfs/spec.py
```diff
@@ -897,6 +897,12 @@ async def _details(
...
+                if not etagVal.startswith('"') and not etagVal.endswith('"'):
+                    data["etag"] = f'"{etagVal}"'
```
nateprewitt: I'm not sure this covers all problematic cases. It's unlikely we'll ever get one of these back from Azure, but probably best to be uniform with how we apply this fix. Values like `1234"` will pass this silently.

A more robust option may be to use `strip("\"")` to remove quotes on the edges and then force requoting each time. Something like the code below.

```suggestion
                data["etag"] = f'"{data["etag"].strip("\"")}"'
```
or if we kept the variable from the walrus operator above:
```suggestion
                etag_val = etag_val.strip("\"")
                data["etag"] = f'"{etag_val}"'
```

adlfs/tests/test_spec.py
```diff
@@ -2571,3 +2571,28 @@ class TestCloseCredential:
...
+    info_etag = fs.info(path, refresh=True)["etag"]
+
+    assert info_etag == ls_etag
+
+    # Validate both etags are quoted
+    assert ls_etag.startswith('"') and info_etag.startswith('"')
+    assert ls_etag.endswith('"') and info_etag.endswith('"')
+
+    # Validate etag from info() is not double quoted
+    assert not info_etag.startswith('""') and not info_etag.endswith('""')
```
nateprewitt: nit; but since we're comparing the etags we can validate one and then the comparison ensures we have everything inline.
```suggestion
    # Validate etag is quoted but only once.
    assert ls_etag.startswith('"') and ls_etag.endswith('"')
    assert not ls_etag.startswith('""') and not ls_etag.endswith('""')
    
    # Validate etags are identical
    assert info_etag == ls_etag
```

adlfs/tests/test_spec.py
```diff
@@ -2571,3 +2571,28 @@ class TestCloseCredential:
...
+
+def test_etag_normalized_form(storage):
+    """
+    Tests a consistent quoted etag format with ls() and info() calls.
+    """
+    fs = AzureBlobFileSystem(
+        account_name=storage.account_name,
+        connection_string=CONN_STR,
+    )
+    path = "data/root/a/file.txt"
```
nateprewitt, line 2585: It looks like we're getting our etag back from Azurite checksumming the file? That seems reasonable for an e2e test, but I'm wondering if we want to unit test the normalization behavior any more for edge cases like an empty value (`""`) or lop-sided quotes to make sure we don't fail unexpectedly. Uniformity is good but we don't want to introduce runtime failures for it.

We could break the normalization into a testable function if needed. I'm not sure if @kyleknap or @anjaliratnam-msft  have an opinion.

nateprewitt (changes requested):
Thanks for the PR, @wonwuakpa-msft, this is a great start! I left some initial feedback on a few things. Please let me know if you think I'm missing something or if you have any questions.

adlfs/spec.py
wonwuakpa-msft: Yea, I can simplify this validation to use just the getter. I agree the service can only send us back either None (when the header isn't set) or a valid Etag string. So, there is no need to handle an empty string.

adlfs/tests/test_spec.py
anjaliratnam-msft: Yeah, I agree. I think it could be worth it to add unit tests just to make sure it's always gets wrapped with double quotes.

adlfs/tests/test_spec.py
```diff
@@ -2571,3 +2571,33 @@ class TestCloseCredential:
...
+
+    assert info_etag == ls_etag
+
+    # Validate both etags are quoted
+    assert ls_etag.startswith('"') and info_etag.startswith('"')
+    assert ls_etag.endswith('"') and info_etag.endswith('"')
+
+    # Validate etag from info() is not double quoted
+    # Validate etag is quoted but only once.
+    assert ls_etag.startswith('"') and ls_etag.endswith('"')
```
anjaliratnam-msft: I think you can get rid of this line since it's already being checked by 2594 and 2595.

adlfs/spec.py
```diff
@@ -1,2337 +1,2363 @@
...
+
+def _create_aio_blob_service_client_from_connection_string(
+    connection_string: str,
+) -> AIOBlobServiceClient:
+    return AIOBlobServiceClient.from_connection_string(
+        conn_str=connection_string,
+        user_agent=_USER_AGENT,
+    )
+
+
```
wonwuakpa-msft: `_strip_quotes_from_etag` and its calling location is the only addition in this file. Not sure if this is the best place to add this helper. It makes the diff really large

nateprewitt:
@wonwuakpa-msft It looks like the latest commit may have modified the newline at the end of spec.py so it's diffing the entire file. Can you take a look at that? Feel free to ping me if you need help.

wonwuakpa-msft:
> @wonwuakpa-msft It looks like the latest commit may have modified the newline at the end of spec.py so it's diffing the entire file. Can you take a look at that? Feel free to ping me if you need help.

Yes, it was changing the line endings of every line in spec.py from LF to CRLF, so git detected that as a new change. Setting the `git config core.autocrlf = input` and changing my settings.json file in VSCode to `/n` resolved it.

adlfs/tests/test_spec.py
```diff
@@ -2571,3 +2572,41 @@ class TestCloseCredential:
...
+    assert ls_etag.startswith('"') and info_etag.startswith('"')
+    assert ls_etag.endswith('"') and info_etag.endswith('"')
+    # Validate etag is not double quoted
+    assert not ls_etag.startswith('""') and not ls_etag.endswith('""')
+
+
+@pytest.mark.parametrize(
+    "input_etag,expected_etag",
+    [
+        pytest.param("0xA123456", '"0xA123456"', id="bug"),
```
anjaliratnam-msft: Instead of doing pytest.param, I think you can just put the values for input_etag and expected_etag, similar to the other tests. I don't really think an id is necessary for these.

anjaliratnam-msft (commented):
This looks good to me, I just had one small comment. Other than that, I think you should be able to rebase off of main as well, since there have been some updates there.

adlfs/spec.py
```diff
@@ -156,6 +156,27 @@ def _create_aio_blob_service_client_from_connection_string(
...
+    Parameters
+    ----------
+    etag: str
+        Raw etag value from Azure
+
+    Returns
+    -------
+    str
+        Returns the normalized etag
+    """
```
nateprewitt: nit; Generally, docstrings aren't needed for private functions unless there's a lot of nuance since end-users won't interact with them.

adlfs/spec.py
```diff
@@ -156,6 +156,27 @@ def _create_aio_blob_service_client_from_connection_string(
     )
 
 
+def _strip_quotes_from_etag(etag: str) -> Optional[str]:
```
nateprewitt: We're quoting the etag so something like this may be clearer about intent.
```suggestion
def _normalize_etag_quotes(etag: str) -> Optional[str]:
```

nateprewitt (commented):
Looks good! A couple minor comments and once @anjaliratnam-msft's comments are addressed we should be set.

adlfs/tests/test_spec.py
```diff
@@ -2571,3 +2572,33 @@ class TestCloseCredential:
...
+    info_etag = fs.info(path, refresh=True)["etag"]
+
+    assert info_etag == ls_etag
+    # Validate both etags are quoted
+    assert ls_etag.startswith('"') and info_etag.startswith('"')
+    assert ls_etag.endswith('"') and info_etag.endswith('"')
+    # Validate etag is not double quoted
+    assert not ls_etag.startswith('""') and not ls_etag.endswith('""')
+
+    assert _normalize_etag_quotes("0xA123456") == '"0xA123456"'
```
anjaliratnam-msft: Oh sorry, I just meant removing the pytest.param part but still keeping it under the parametrization under a separate test. something like this
```suggestion
    @pytest.mark.parametrize(
    "input_etag,expected_etag",
    [
        ("0xA123456", '"0xA123456"')
        ('"0xA123456"', '"0xA123456"'),
        ('""0xA123456""', '"0xA123456"'),
        ('"0xA123456', '"0xA123456"'),
        ('0xA123456"', '"0xA123456"'),
        (None, None),
    ],
```

nateprewitt (approved):
:shipit: Thank you @wonwuakpa-msft!

kyleknap (approved):
Looks good to me! :ship:

PR #545, opened 2026-04-23
Removed ADLS Gen1 Support

### LINKED ISSUE #533: Remove support for Azure Data Lake Storage (ADLS) Gen 1
Tracking issue for removing support for ADLS Gen 1. PR for emitting warning for imminent dropping of support has been merged: https://github.com/fsspec/adlfs/pull/529. See PR for more details on plan.

anjaliratnam-msft:
We are removing support for ADLS Gen1. Users will no longer be able to use the `AzureDatalakeFileSystem` class and it is recommended to use the `az://` protocol or the `AzureBlobFileSystem` class which support both Azure Blob Storage and Azure Data Lake Storage Gen2. If users attempt to use the `AzureDatalakeFileSystem` class, they will see the following:

> RuntimeError: Azure Data Lake Storage (ADLS) Gen1 has been retired since February 29, 2024 and is no longer supported by adlfs. It is recommended to use the az:// protocol and/or adlfs.AzureBlobFileSystem class which support Azure Blob Storage and Azure Data Lake Storage Gen2.

This resolves #533.


adlfs/tests/test_init.py
```diff
@@ -0,0 +1,40 @@
...
+from adlfs import AzureBlobFile, AzureBlobFileSystem
+
+
+def test_adl_protocol_attribute_error():
+    with pytest.raises(
+        AttributeError,
+        match="Gen1 has been retired since February 29, 2024 and "
+        "is no longer supported by adlfs",
+    ):
+        fsspec.filesystem("adl")
```
kyleknap, line 13: I like that a test case was added using the the fsspec interfaces. Could we add additional test cases for `fsspec.get_filesystem_class()` and `fsspec.open()` as well? I think those are the main entry points we'd want this error message propagated.

adlfs/tests/test_init.py
```diff
@@ -0,0 +1,40 @@
+import fsspec
+import pytest
+
+from adlfs import AzureBlobFile, AzureBlobFileSystem
+
+
+def test_adl_protocol_attribute_error():
+    with pytest.raises(
+        AttributeError,
+        match="Gen1 has been retired since February 29, 2024 and "
```
kyleknap: If we find ourselves needing to resuse this string, it might make sense to make it a fixture or module level constant to make it easier to update the tests if we ever need to change the messaging.

kyleknap (commented):
This looks good. I like the approach that we ended up with here. I just had a couple of suggestions on the tests but that was about it.

kyleknap (approved):
Looks good! :ship:

potiuk:
Cool. Looking forward to release to unblock some of our dependencies in Airflow

nateprewitt:
@potiuk [2026.5.0](https://pypi.org/project/adlfs/2026.5.0/) should be available on PyPI now.

PR #549, opened 2026-06-08
Fixed ls() bug when files have the same prefix

### LINKED ISSUE #406: Cannot read parquet from Azure when archive file exists
This is the setup in the blob storage:
```
container-name/YEAR2023/MONTH03/DAY15/file_name.parquet
container-name/YEAR2023/MONTH03/DAY15/file_name.parquet.archive_202303162004
container-name/YEAR2023/MONTH03/DAY14/file_name.parquet
```
I get the following error: 
`azure.core.exceptions.HttpResponseError: The specifed resource name contains invalid characters.`
When I am trying to read:
`file_name = "az://container-name/YEAR2023/MONTH03/DAY15/file_name.parquet"`
I only get the error when the following file exists:
`"az://container-name/YEAR2023/MONTH03/DAY15/file_name.parquet.archive_202303162004"`

`file_name = "az://container-name/YEAR2023/MONTH03/DAY14/file_name.parquet"`
is read fine.

There appears to be a regression from `2022.10.0` to any newer version as with `2022.10.0` the below works:

```python
import pandas as pd
df = pd.read_parquet('az://container-name/YEAR2023/MONTH03/DAY15/file_name.parquet',storage_options={'connection_string': con_str, })

```

works with:
```
adlfs == 2022.10.0
fsspec ==  2023.3.0
```

but not newer versions of adlfs

With the newest version of adlfs this works though:

```
import fsspec
import pandas as pd
fs = fsspec.filesystem('az', connection_string=con_str)
fs.ls("/container-name/YEAR2023/MONTH03/DAY15/")
df = pd.read_parquet(file_name, storage_options={'connection_string': con_str})
```

but not if i ran the command 
`df = pd.read_parquet('az://container-name/YEAR2023/MONTH03/DAY15/file_name.parquet',storage_options={'connection_string': con_str})` first.

### LINKED ISSUE #409: `AzureBlobFileSystem.ls(file_path)` does not work if other files having the same name + additional suffix exist in the same directory
## Versions:
* `adlfs==2023.1.0`
* `fsspec==2023.3.0`

(This problem seemingly started from `adlfs==2022.11.0`)

## Summary

Suppose we have a file `X` and other files `X.1`, `X.2` etc. having the same name plus additional suffix in the same directory. For example:
```
path/
  to/
    dummy.txt
    dummy.txt.1
    dummy.txt.2
    ...
```

In this case, `AzureBlobFileSystem.ls("container_name/path/to/dummy.txt")` returns an empty result.

## POC

```
from adlfs.spec import AzureBlobFileSystem
from azure.identity.aio import AzureCliCredential

fs = AzureBlobFileSystem(account_name="my_account", credential=AzureCliCredential())

fs.ls("my_container/path/to/", invalidate_cache=True)
> ['my_container/path/to/dummy.txt']

fs.ls("my_container/path/to/dummy.txt", invalidate_cache=True)
> ['my_container/path/to/dummy.txt']
```

Now we create an empty file `dummy.txt.1` in the same directory. Then `AzureBlobFileSystem` starts returning inconsistent results.

```
fs.ls("my_container/path/to/", invalidate_cache=True)
> ['my_container/path/to/dummy.txt', 'my_container/path/to/dummy.txt.1']   # OK

fs.ls("my_container/path/to/dummy.txt", invalidate_cache=True)
> []  # KO

fs.ls("my_container/path/to/dummy.txt.1", invalidate_cache=True)
> ['my_container/path/to/dummy.txt.1']  # OK
```

This problem is probably linked to #406.

anjaliratnam-msft:
This addresses a bug where a filepath is passed into `ls()` and it shares the same prefix as other files in the same directory resulting in incorrect results. This resolves #406 and resolves #409.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Unreleased
 ----------
-- .
+- Fixed `ls()` returning incorrect results when a file shares the same prefix as other files in the same directory (#409)
 
 2026.5.0
 --------
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -946,13 +946,16 @@ class AzureBlobFileSystem(AsyncFileSystem):
                 data["name"] = data["name"].rstrip("/")
             output.append(data)
         if target_path:
-            if (
-                len(output) == 1
-                and output[0]["type"] == "file"
-                and not self.version_aware
-            ):
-                # This handles the case where path is a file passed to ls
-                return output
+            if not self.version_aware:
+                # This handles the case where a file path is passed to ls and there are
+                # multiple files with the same prefix.
+                file = [
+                    o
+                    for o in output
+                    if o["name"] == target_path and o["type"] == "file"
+                ]
+                if file:
+                    return file
             output = await filter_blobs(
                 output,
                 target_path,
--- a/adlfs/tests/test_spec.py
+++ b/adlfs/tests/test_spec.py
@@ -2610,3 +2610,20 @@ def test_etag_normalized_form(storage):
 )
 def test_striping_etag(input_etag, expected_etag):
     assert _normalize_etag_quotes(input_etag) == expected_etag
+
+
+def test_ls_files_with_same_prefix(storage):
+    fs = AzureBlobFileSystem(
+        account_name=storage.account_name,
+        connection_string=CONN_STR,
+    )
+    path1 = "data/test/file.txt"
+    path2 = "data/test/file.txt.1"
+    fs.touch(path1)
+    fs.touch(path2)
+
+    assert fs.ls("data/test/") == [path1, path2]
+    assert fs.ls("data/test/file.txt") == [path1]
+    assert fs.ls("data/test/file.txt.1") == [path2]
+
+    fs.rm("data/test", recursive=True)


martindurant:
@shcheklein , maybe?

@anjaliratnam-msft , was there a reason this stalled?

anjaliratnam-msft:
Hi @martindurant. Yeah, after taking a look at it again and talking to some team members, there's some changes I want to make to the PR today. Mainly just moving the logic to `filter_blobs`. After that, someone from my team should be able to review it next week.

martindurant:
Perfect, thank you.

adlfs/utils.py
```diff
@@ -38,13 +38,27 @@ async def filter_blobs(
...
-    target_path = target_path.strip(" " + delimiter) + delimiter
+    # remove delimiter and spaces
+    target_path = target_path.strip(" " + delimiter)
+
+    # If target_path is itself a file, return just that file. This handles
+    # ls("container/path/to/file") where other blobs share the same prefix.
+    exact_file = [
+        b
+        for b in blobs
+        if b["name"].strip(" " + delimiter) == target_path
```
kyleknap: Can we instead match exactly against the `target_path` prior to any modification? e.g.,
```
if b["name"] == target_path
```
and then apply the modifications to `target_path` to build the `prefix` later on?

Mainly, it makes the code a little more straightforward to follow that we are looking for an exact match with no modifications.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Unreleased
 ----------
 - Added a `content_settings` parameter to `open()` (write modes), `pipe_file()` and `put_file()` to set blob content settings (content type, content disposition, cache control, ...) on write. Accepts a `dict` (recommended) or an `azure.storage.blob.ContentSettings` instance. [#554](https://github.com/fsspec/adlfs/pull/554)
+- Fixed `ls()` returning incorrect results when a file shares the same prefix as other files in the same directory (#409)
```
kyleknap: Can we turn `#409` into a hyper-link similar to the entry above? Right now in the rendered markdown, no link gets rendered so I think we need to provide the explicit URL (unlike in GitHub PRs/issues).

nateprewitt: `fsspec/adlfs#409` should render correctly so you're not having to do the markdown link.

anjaliratnam-msft: This didn't seem to work so I just left the URL.

nateprewitt: That's totally fine. Just so I know before recommending again, where was it not working? It looks like it works fine here: https://github.com/nateprewitt/adlfs/blob/2f470b4208d9575a66d7156c3774ca2cdb3b48c6/CHANGELOG.md

adlfs/utils.py
```diff
@@ -38,13 +38,27 @@ async def filter_blobs(
 
     version_id: Spefic blob version ID to be returned
     """
-    # remove delimiter and spaces, then add delimiter at the end
-    target_path = target_path.strip(" " + delimiter) + delimiter
+    # remove delimiter and spaces
+    target_path = target_path.strip(" " + delimiter)
```
kyleknap: Should we lower this as well to after the exact file logic? Specifically, I was thinking we'd make the line after determining whether an `exact_file` was provided would be:
```
prefix = target_path.strip(" " + delimiter) + delimiter
```
That way we would only be comparing exact filenames based on the exact `target_path` provided.

CHANGELOG.md
anjaliratnam-msft: Hmm interesting, maybe it just doesn't show up in the preview.

PR #551, opened 2026-06-15
Remove sync credentials

jamesmyatt:
The sync credentials were only created by internal methods, and only used in AzureBlobFile.connect_client, when the async credentials should have been preferred anyway.

Part of refactoring of `do_connect`-related logic: https://github.com/fsspec/adlfs/pull/504

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -362,12 +362,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
             and self.client_id is not None
             and self.connection_string is None
         ):
-            (
-                self.credential,
-                self.sync_credential,
-            ) = self._get_credential_from_service_principal()
-        else:
-            self.sync_credential = None
+            self.credential = self._get_credential_from_service_principal()
 
         # Solving issue in https://github.com/fsspec/adlfs/issues/270
         if (
@@ -377,10 +372,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
             and self.account_key is None
             and self.connection_string is None
         ):
-            (
-                self.credential,
-                self.sync_credential,
-            ) = self._get_default_azure_credential(**kwargs)
+            self.credential = self._get_default_azure_credential(**kwargs)
 
         self.do_connect()
         weakref.finalize(self, sync, self.loop, close_service_client, self)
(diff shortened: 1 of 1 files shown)


adlfs/spec.py
```diff
@@ -2140,7 +2116,7 @@ def connect_client(self):
                     f"https://{self.fs.account_name}.blob.core.windows.net"
                 )
 
-            creds = [self.fs.sync_credential, self.fs.account_key, self.fs.credential]
+            creds = [self.fs.credential, self.fs.account_key]
```
kyleknap, line 2119: Not actionable but just calling this out... Technically this is a change of behavior as before if a `credential` was provided to the file system a `sync_credential` would not be set and `account_key` would thus take precedence over the explicitly provided `credential`. That being said, this update seems reasonable as:

1. It make it consistent with how the filesystem class resolves credentials (e.g. it prioritizes `self.credential` over `self.account_key` no matter what)
2. In most cases, the container client will already be instantiated from the file system so this code path in general should not be hit unless `connect_client` is explicitly called.

adlfs/spec.py
```diff
@@ -362,12 +362,7 @@ def __init__(
             and self.client_id is not None
             and self.connection_string is None
         ):
-            (
-                self.credential,
-                self.sync_credential,
```
kyleknap, line 367: Could we add a changelog entry noting that we removed the `sync_credential` property as it was unused and label it as `**Breaking**` similar to other entries?

kyleknap (commented):
Hi @jamesmyatt thanks for following up on refactoring the credential resolution. Even though it is technically a breaking change to remove the public `sync_credential` property, I'm in favor of the change (as you pointed out) sync credentials are no longer used and the project only [documents usage of async credentials](https://github.com/fsspec/adlfs/blob/8a1246830955b342ce1b9272162c86f83ce5ef6b/adlfs/spec.py#L193). So that will help clear up some confusion. Furthermore, previous maintainers were in favor of removing the sync credentials completely once the implementation no longer used sync creds: https://github.com/fsspec/adlfs/pull/132

Overall, from this first pass, the change looks good. For now, I just had a small comment about adding a changelog entry for it. Otherwise, I plan to pull down the change later this week to do a deeper review on it.

Issue #515, opened 2025-10-16
Document Microsoft OneLake support


kevinjqliu:
Context: https://github.com/fsspec/adlfs/issues/416#issuecomment-3408577353

Document support for OneLake, including
* OneLake Blob endpoint, `onelake.blob.fabric.microsoft.com`. (Azure Blob Storage -compatible)
* OneLake DFS endpoint, `onelake.dfs.fabric.microsoft.com`. (Azure Data Lake Storage -compatible)

Related issues:
* #416
* #486
* #401
* #513

kevinjqliu:
https://github.com/fsspec/adlfs/pull/513#issuecomment-3408715758

OneLake blob endpoint is supported
OneLake dfs endpoint is not yet supported, blocked on #401

hutch3232:
There are a couple stale issues and even a PR related to Fabric OneLake. What can potential contributors do to help push this forward?

PR #552, opened 2026-06-15
Updated documentation for setting different endpoints

### LINKED ISSUE #402: Maybe fsspec can't working well in Azure CN
I found that whe the Protocol is abfs[s]
In this source file:adlfs/spec.py
STORE_SUFFIX = ".dfs.core.windows.net"
It should working well in Azure,
But in Azure CN,it should be STORE_SUFFIX = "dfs.core.chinacloudapi.cn"
And other like f"https://{self.account_name}.blob.core.windows.net"
in Azure CN should be f"https://{self.account_name}.blob.core.chinacloudapi.cn"
Or here:
 r"(?P<account_name>.+)\.(dfs|blob)\.core\.windows\.net", host
in Azure CN should be  r"(?P<account_name>.+)\.(dfs|blob)\.core\.chinacloudapi\.cn", host

anjaliratnam-msft:
This PR updates the documentation to instruct users how to set different endpoints. This resolves #402.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Unreleased
 ----------
-- Fixed a bug in `_strip_protocol` and `_get_kwargs_from_urls` to correctly handle Azure sovereign cloud URLs (#402)
+- .
 
 2026.5.0
 --------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Unreleased
 ----------
-- .
+- Fixed a bug in `_strip_protocol` and `_get_kwargs_from_urls` to correctly handle Azure sovereign cloud URLs (#402)
 
 2026.5.0
 --------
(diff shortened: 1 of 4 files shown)


kyleknap:
Hey @anjaliratnam-msft I'm thinking we hold off updating any of the URL parsing for now as end users can set a different endpoint, they just have to use the `account_url` and `account_name` parameters in tandem. I put an example of it in the original GitHub issue: https://github.com/fsspec/adlfs/issues/402#issuecomment-4724184969

Mainly, to my understanding the `abfs://<file_system>@<account_name>.dfs.core.windows.net/<path>/<file_name>` format was specifically to handle the ABFS driver style of URIs and I'm not sure if it supports URIs outside of `dfs.core.windows.net` per the documentation: https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri#uri-syntax. So, I'm hesitant to start supporting different URIs in adlfs especially if the ABFS driver does not support these URIs and because there is already a way to specify different endpoint in adlfs via parameters.

I'm thinking maybe we can repurpose this PR to update the documentation on how to connect to different endpoints? Support of Microsoft OneLake is a similar story (where a different endpoint needs to be set to use with adlfs): https://github.com/fsspec/adlfs/issues/515 So maybe we can include that in the doc update as well? Thoughts?

PR #554, opened 2026-07-06
Forward content_settings on write (open, pipe_file, put_file)

shcheklein:
Note: AI was heavily used to generate this. I reviewed the design principle and discussions on this topic across fsspec implementations.

## Problem

`AzureBlobFile` only forwards `metadata` when committing a write — it drops content settings entirely. So there is no way to set `content_type` / `content_disposition` / `cache_control` / `content_encoding` when writing a blob through `fs.open(..., "wb")`, `pipe_file`, or `put_file`, even though the Azure SDK's `commit_block_list` / `upload_blob` / `create_append_blob` accept `content_settings`. This makes adlfs inconsistent with `s3fs` (extra kwargs → `s3_additional_kwargs`) and `gcsfs` (`content_type` / `fixed_key_metadata`), which set these inline in the atomic write.

## Change

Add a `content_settings` parameter to `open()` (write modes), `pipe_file()`, and `put_file()`, applied **atomically** as part of the write commit (block, empty, and append blobs) — not a separate follow-up request.

To keep the Azure SDK out of the caller's imports, the documented/typed form is a plain `dict`; an `azure.storage.blob.ContentSettings` instance is also accepted for convenience.

```python
import fsspec

with fsspec.open(
    "abfs://cont/report.pdf",
    mode="wb",
    content_settings={
        "content_type": "application/pdf",
        "content_disposition": 'attachment; filename="report.pdf"',
        "cache_control": "max-age=3600",
    },
    account_name="myaccount",
    connection_string=CONN_STR,
) as f:
    f.write(data)

# also works with pipe_file / put_file
fs.pipe_file("cont/x.pdf", data, content_settings={"content_type": "application/pdf"})
```

## Notes

- Scoped to content settings only. An earlier revision also let `pipe_file`/`put_file` override the hard-coded `metadata={"is_directory": "false"}`; that's unrelated to content settings (and metadata is already settable via `fs.open(metadata=...)`), so it was reverted.
- Deferred, per discussion: no top-level `content_type=` kwarg for now — better introduced once the fsspec implementations converge on a shared interface (see [filesystem_spec#916](https://github.com/fsspec/filesystem_spec/issues/916)); a top-level field can be added alongside `content_settings` later.

## Tests

Per-method tests (verified against azurite):
- `test_content_settings_open` — streaming `open("wb")` + empty write
- `test_content_settings_append` — `open("ab")` (append blob)
- `test_content_settings_pipe_file` — `pipe_file`
- `test_content_settings_put_file` — `put_file`
- `test_content_settings_accepts_object` — a `ContentSettings` instance is accepted

Each verifies the settings round-trip via `fs.info(...)["content_settings"]`. Existing `test_pipe_file_timeout` / `test_put_file_timeout` updated for the new `content_settings` kwarg.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -1519,7 +1519,6 @@ class AzureBlobFileSystem(AsyncFileSystem):
         """Set the bytes of given file"""
         if kwargs.pop("mode", "") == "create":
             overwrite = False
-        metadata = kwargs.pop("metadata", None) or {"is_directory": "false"}
         content_settings = _as_content_settings(kwargs.pop("content_settings", None))
         container_name, path, _ = self.split_path(path)
         async with self.service_client.get_blob_client(
@@ -1529,7 +1528,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
                 result = await bc.upload_blob(
                     data=value,
                     overwrite=overwrite,
-                    metadata=metadata,
+                    metadata={"is_directory": "false"},
                     content_settings=content_settings,
                     max_concurrency=max_concurrency or self.max_concurrency,
                     **self._timeout_kwargs,
@@ -1761,7 +1760,6 @@ class AzureBlobFileSystem(AsyncFileSystem):
 
         if kwargs.pop("mode", "") == "create":
             overwrite = False
-        metadata = kwargs.pop("metadata", None) or {"is_directory": "false"}
         content_settings = _as_content_settings(kwargs.pop("content_settings", None))
         container_name, path, _ = self.split_path(rpath, delimiter=delimiter)
 
@@ -1776,7 +1774,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
                         await bc.upload_blob(
                             f1,
                             overwrite=overwrite,
-                            metadata=metadata,
+                            metadata={"is_directory": "false"},
                             content_settings=content_settings,
                             raw_response_hook=make_callback(
                                 "upload_stream_current", callback
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -164,6 +164,19 @@ def _normalize_etag_quotes(etag: str) -> Optional[str]:
     return f'"{etag.strip(double_quote)}"'
 
 
+def _as_content_settings(
+    content_settings: Optional[Union[ContentSettings, dict]],
+) -> Optional[ContentSettings]:
+    """Coerce ``content_settings`` given as a dict into a ``ContentSettings``.
+
+    Accepts either a plain ``dict`` (recommended — keeps the Azure SDK out of the
+    caller's imports) or a ``ContentSettings`` instance.
+    """
+    if content_settings is None or isinstance(content_settings, ContentSettings):
+        return content_settings
+    return ContentSettings(**content_settings)
+
+
 class AzureBlobFileSystem(AsyncFileSystem):
     """
     Access Azure Datalake Gen2 and Azure Storage if it were a file system using Multiprotocol Access
@@ -1507,6 +1520,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         if kwargs.pop("mode", "") == "create":
             overwrite = False
         metadata = kwargs.pop("metadata", None) or {"is_directory": "false"}
+        content_settings = _as_content_settings(kwargs.pop("content_settings", None))
         container_name, path, _ = self.split_path(path)
         async with self.service_client.get_blob_client(
             container=container_name, blob=path
@@ -1516,6 +1530,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
                     data=value,
                     overwrite=overwrite,
                     metadata=metadata,
+                    content_settings=content_settings,
                     max_concurrency=max_concurrency or self.max_concurrency,
                     **self._timeout_kwargs,
                     **kwargs,
@@ -1747,6 +1762,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         if kwargs.pop("mode", "") == "create":
             overwrite = False
         metadata = kwargs.pop("metadata", None) or {"is_directory": "false"}
+        content_settings = _as_content_settings(kwargs.pop("content_settings", None))
         container_name, path, _ = self.split_path(rpath, delimiter=delimiter)
 
         if os.path.isdir(lpath):
@@ -1761,6 +1777,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
                             f1,
                             overwrite=overwrite,
                             metadata=metadata,
+                            content_settings=content_settings,
                             raw_response_hook=make_callback(
                                 "upload_stream_current", callback
                             ),
@@ -1902,7 +1919,7 @@ class AzureBlobFileSystem(AsyncFileSystem):
         cache_type="readahead",
         metadata=None,
         version_id: Optional[str] = None,
-        content_settings: Optional[ContentSettings] = None,
+        content_settings: Optional[Union[ContentSettings, dict]] = None,
         **kwargs,
     ):
         """Open a file on the datalake, or a block blob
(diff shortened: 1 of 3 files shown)


kyleknap:
@shcheklein Thanks for the pull request! I think both exposing the content setting related values and allowing overriding of metadata across all write methods make sense functionally.

It's unfortunate that `s3fs` and `gcsfs` already diverge in being able to set content type, encoding, etc. I think if they were consistent, it would probably make sense to match their interfaces. So, matching SDK interfaces seems reasonable at this point.

We'll take a closer review at the PR later this week so we can make progress toward getting this merged in.

shcheklein:
@kyleknap yes, agreed, I even found a few discussions:

https://github.com/fsspec/filesystem_spec/issues/916

I was looking also into making this in a way that others can pick it up.

One practical thing we can do is to add at least `content_type=` (similar to what GCS has now). And then when all of the specs decide on the generic "additional_params" we can convert content_settings into that one and drop eventually. For now S3 and GCS are way too specific also, so hard to unify, and I'm not sure it makes sense making this decision here.

adlfs/spec.py
```diff
@@ -2073,6 +2087,7 @@ def __init__(
 
         else:
             self._metadata = metadata or {"is_directory": "false"}
+            self.content_settings = content_settings
```
kyleknap: I'm thinking let's make this property private for now (e.g., `self._content_settings`) unless we need to expose it publicly. Mainly:

1. If we exposed publicly, we probably should fill the property when the file-like object is opened in read mode
2. If we ever unify the `content_type`, `content_encoding` parameters in the future, I'd suspect similar properties would be exposed on the file-like object interface so we'd have more flexibility here if `content_settings` is kept private for now.

adlfs/tests/test_spec.py
```diff
@@ -1413,6 +1413,57 @@ def test_metadata_write(storage):
...
+    )
+    fs.mkdir("test-content-settings")
+    data = b"0123456789"
+    settings = ContentSettings(
+        content_type="application/pdf",
+        content_disposition='attachment; filename="f.pdf"',
+        cache_control="max-age=3600",
+    )
+
+    # streaming write (block blob)
```
kyleknap: Would it be possible to break up each method into separate test methods? I realize this is just following the pattern used by previous tests, but breaking into separate test methods will make it easier to track which adlfs methods have coverage and also easier to know which methods are impacted by changes in the future when running tests.

adlfs/tests/test_spec.py
```diff
@@ -1413,6 +1413,57 @@ def test_metadata_write(storage):
...
+    ) as f:
+        f.write(data)
+    cs = fs.info("test-content-settings/file.pdf")["content_settings"]
+    assert cs["content_type"] == "application/pdf"
+    assert cs["content_disposition"] == 'attachment; filename="f.pdf"'
+    assert cs["cache_control"] == "max-age=3600"
+
+    # empty streaming write still applies content settings
+    with fs.open(
+        "test-content-settings/empty.pdf", "wb", content_settings=settings
```
kyleknap: Could we also add a test case for `ab` mode to make sure content settings are used when initializing an append blob?

adlfs/tests/test_spec.py
```diff
@@ -1413,6 +1413,57 @@ def test_metadata_write(storage):
...
+    ) as f:
+        f.write(b"")
+    assert (
+        fs.info("test-content-settings/empty.pdf")["content_settings"]["content_type"]
+        == "application/pdf"
+    )
+
+    # pipe_file carries content settings and custom metadata together
+    fs.pipe_file(
```
kyleknap, line 1467: Can we also make sure we have a test for setting metadata for `put_file()`?

adlfs/tests/test_spec.py
```diff
@@ -1413,6 +1413,57 @@ def test_metadata_write(storage):
     fs.rmdir("test-metadata-write")
 
 
+def test_content_settings_write(storage):
```
kyleknap: Let's also make sure to add changelog entries for both the new content_setting parameter and being able to override metadata in `put_file()` and `pipe_file()`.

adlfs/spec.py
```diff
@@ -1899,6 +1902,7 @@ def _open(
         cache_type="readahead",
         metadata=None,
         version_id: Optional[str] = None,
+        content_settings: Optional[ContentSettings] = None,
```
kyleknap: In terms of support content settings, I'm thinking instead of publicly advertising to use Azure SDK's `ContentSettings` type, we just make the documented type here a `Dict[str, str]` similar to `metadata`. My main concern in documenting to use the `ContentSettings` type is that now downstream users would have to import the SDK when adlfs is intended to be generic/injectable. For example, if we exposed it as dictionary it would be:
```python
import fsspec

storage_options = {"account_name": "myaccount", "connection_string": CONN_STR}

with fsspec.open(
    "abfs://container/file.pdf",
    mode="wb",
    content_settings={
        "content_type": "application/pdf",
        "content_disposition": 'attachment; filename="f.pdf"',
        "cache_control": "max-age=3600",
    },
    **storage_options,
) as f:
    f.write(b"0123456789")
```
vs needing to import the SDK type:
```python
import fsspec
from azure.storage.blob import ContentSettings

storage_options = {"account_name": "myaccount", "connection_string": CONN_STR}

settings =
with fsspec.open(
    "abfs://container/file.pdf",
    mode="wb",
    content_settings= ContentSettings(
        content_type="application/pdf",
        content_disposition='attachment; filename="f.pdf"',
        cache_control="max-age=3600",
    ),
    **storage_options,
) as f:
    f.write(b"0123456789")
```
Also, I think this makes it a cleaner story for moving these value out of `content_settings` if we ever unify these content configurations.

This being said, we should still make sure to have compatibility (and a test case) for the actual `ContentSettings` type in `open()`, `put_file()`, and `pipe_file()` for backwards compatibility, but we'd just not include it in the typing or documentation and push users to use the dictionary form.

Thoughts? Happy to expand on this more.

kyleknap (commented):
Thanks @shcheklein for the additional details! Yeah I agree that we should hold off adding `content_type` till we know we are likely to unify these parameters. Furthermore, it should not be an issue in supporting both a top-level `content_type` and one nested in `content_settings` at some point in the future.

I just left some suggestions on the PR. Overall, it looked good. Probably the biggest suggestion was that we tweak the `content_settings` parameter to also support providing these values as a dictionary to avoid forcing people from needing to import the SDK. We'd also document/recommend to users to use this dictionary format but we'd also accept (but not necessarily advertise) the `ContentSettings` object for compatibility use cases.

adlfs/spec.py
shcheklein: 👍 making it private

adlfs/tests/test_spec.py
shcheklein: yep, added below

adlfs/tests/test_spec.py
shcheklein: I've actually removed this - I think this is not relevant to this PR (also removed from tests the ability to override metadata). It should be probably a separate PR

adlfs/spec.py
shcheklein: yep, good idea. I'm not sure about the point `for backwards compatibility`, but we can support both. Probably makes sense since file info returns it (?).

shcheklein: Fixed this

adlfs/tests/test_spec.py
shcheklein: Yep. adeed

adlfs/tests/test_spec.py
shcheklein: yes, I was trying to adhere to the style in the file (tests looked monolithic), but yes, it is better to have them smaller

shcheklein:
@kyleknap thank for the quick review! I've done another pass, cleaned things a bit and added more tests

adlfs/tests/test_spec.py
kyleknap: Sounds good. Happy to review it if you send a separate PR for it.

adlfs/spec.py
kyleknap: > Probably makes sense since file info returns it (?).

Yeah the two reasons why I thought we'd still want to accept it was:

1. It was already supported for some methods (`put_file` and `pipe_file`) so removing it would break people
2. It is returned in the `info()` so technically you could try to roundtrip the `ContentSettings` object back to a write.

kyleknap (approved):
This looks good to me. :ship: Thanks @shcheklein!

PR #555, opened 2026-07-24
Accept and propagate `api_version` argument through factory functions

amotl:
Hi. Thanks a stack for conceiving adlfs. When aiming to run integration tests (https://github.com/panodata/omniload/pull/249), we ran into a problem.

## Problem

The [Azurite](https://github.com/Azure/Azurite) service emulator does not always implement the newest API versions of the original Azure Blob and Data Lake Storage APIs running in production.

```python
azure.core.exceptions.HttpResponseError: The API version 2026-06-06 is not supported by Azurite.
Please upgrade Azurite to latest version and retry. If you are using Azurite in Visual Studio,
please check you have installed latest Visual Studio patch. Azurite command line parameter
"--skipApiVersionCheck" or Visual Studio Code configuration "Skip Api Version Check" can skip this error.
```

## Solution

As of today, it is needed to define `api_version=2025-11-05` to connect successfully to Azurite. This patch allows users to define the API version when using adlfs.

### DIFF
--- a/adlfs/spec.py
+++ b/adlfs/spec.py
@@ -136,9 +136,11 @@ def _create_aio_blob_service_client(
     account_url: str,
     location_mode: Optional[str] = None,
     credential: CredentialType | None = None,
+    api_version: Optional[str] = None,
 ) -> AIOBlobServiceClient:
     service_client_kwargs = {
         "account_url": account_url,
+        "api_version": api_version,
         "user_agent": _USER_AGENT,
     }
     if credential is not None:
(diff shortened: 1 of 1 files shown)


anjaliratnam-msft:
Hi @amotl! Before doing a deeper review, we just want to make sure that this change is necessary. Is there a reason you're unable to use the `--skipApiVersionCheck` flag as mentioned in the error message? This should already handle the mismatched API version without needing to make any changes.

amotl:
Ah, that flag is on the Azurite container, right. I will try, thanks!

